PR #5, opened 2020-01-27
Feature/GitHub actions

sinergise-anze:
This MR removes GitLab CI instructions and replaces them with GitHub Actions workflows.


.github/workflows/on_release.yaml
```diff
@@ -0,0 +1,44 @@
...
+          npm install
+
+      - name: Build and deploy
+        if: success() && startsWith(github.ref, 'refs/tags/v')
+        env:
+          GITHUB_REF_TAG: ${{ github.ref }}
+          NPMJS_REGISTRY: registry.npmjs.org
+          NPMJS_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+          export VERSION=${GITHUB_REF_TAG:10:50}
```
zcernigoj: Can't find it in the docs how the `github.ref` looks like - is the version part always in the same place in the string (does the string change (apart from the version part of it)?

sinergise-anze: It always starts with `refs/tags/`, though I'm also having trouble locating appropriate docs.

I agree it is not obvious what is happening... I have changed the extraction of `VERSION` a bit, do you think this is better?

zcernigoj: Yes, I think it's better now.
Resolving.

PR #6, opened 2020-01-27
makeLayers add switch to create a specific layer object

zcernigoj:
This MR makes the creation of layers in `LayersFactory`'s method `makeLayers` more specific

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -18,6 +18,7 @@ import { S2L1CLayer } from 'src/layer/S2L1CLayer';
 import { MODISLayer } from 'src/layer/MODISLayer';
 import { DEMLayer } from 'src/layer/DEMLayer';
 import { Landsat8AWSLayer } from 'src/layer/Landsat8AWSLayer';
+import { Polarization } from 'src';
 
 type GetCapabilitiesXml = {
   WMS_Capabilities: {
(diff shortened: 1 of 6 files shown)


src/layer/LayersFactory.ts
```diff
@@ -160,15 +161,86 @@ export class LayersFactory {
...
+          case DATASET_AWS_S1GRD_IW.id:
+            return new S1GRDIWAWSLayer(
+              LayersFactory.parseSHInstanceId(baseUrl),
+              layerId,
+              null,
+              null,
+              null,
+              title,
+              description,
+              Polarization.SV,
```
zcernigoj: We can't assume that polarization is SV but we can't get the data from the layer either (except if we strictly limit the `layerId` options so that it must contain the polarization). Do you have any other idea?

sinergise-anze: If user is authenticated (`setAuthToken` was called), we might be able to get this information from `/layers/` endpoint. However this is only needed in case `getMap` was called with Processing API - WMS doesn't need it as it is specified on the layer anyway. I would suggest fetching it lazily, similar to what we do with `backscatterCoeff` and `orthorectify` parameters. Note that they are not used by Processing API (but should be).

sinergise-anze: As discussed live, I have rearranged Processing API support a bit. Dataset specific layers can now update the Processing API payload as they see fit, with the parameters they fetch (lazily) from the `/layers/` endpoint.

I have added the Storybook for S-1, but it depends on the appropriate layer being available. Not really useful unless the user is willing to dig in and test different configurations. The layer name is also hardcoded to `S1GRDIWDV` (I don't think making it configurable would be much of a help to users). In other words, we need to find a better way to configure storybook fixtures.

Can you maybe take a look at the changes and comment on them?

stories/index.stories.js
```diff
@@ -123,6 +123,37 @@ export const S2GetMapProcessing = () => {
...
+  }
+  setAuthToken(process.env.STORYBOOK_AUTH_TOKEN);
+
+  const img = document.createElement('img');
+  img.width = '512';
+  img.height = '512';
+
+  // getMap is async:
+  const perform = async () => {
+    const layer = new S1GRDIWAWSLayer(instanceId, 'S1GRDIWDV');
```
zcernigoj: Can't make this example work even after creating the layer with this name and trying all possible polarizations and scripts.

sinergise-anze: Yes, this is a problem of V2/V3 scripts and dataproducts. The dataProduct you have selected is not compatible with Processing API (yet - I assume), so you need to enter an evalscript. Note that it must be V3, for example:

```
//VERSION=3
function setup() {
  return {
    input: ["VV"],
    output: { bands: 3 }
  };
}

function evaluatePixel(sample) {
  return [
    2 * sample.VV,
    2 * sample.VV,
    2 * sample.VV,
  ];
}
```

Note that polarization needs to be `DV` (`VV+VH`) for this storybook to work.

zcernigoj: Thanks for the answer. Your evalscript works. In was indeed the problem with the version.

src/layer/LayersFactory.ts
zcernigoj: Changes look good and make sense.

I tried without `Polarization.SV` here in `makeLayers()` and it works, so I think we don't need the switch (can be as it was before this MR).

sinergise-anze: Thanks, good point! Since we don't supply `evalscript` at this step either, there is no reason to supply polarization (WMS knows it, Processing needs to ask service about other things anyway). I have removed the special initialization of S1 layer now.

zcernigoj: Nice. Resolving.

zcernigoj: Nice. Resolving.

PR #8, opened 2020-01-29
Feature/examples get auth token

zcernigoj:
Update examples and get auth token automatically in the examples so changing the auth token in env manually is not needed.


README.md
```diff
@@ -162,11 +167,79 @@ If we already have a WMS GetMap URL, we can use it directly:
...
+let xhr = new XMLHttpRequest();
+xhr.open("POST", 'https://services.sentinel-hub.com/oauth/token', true);
+xhr.setRequestHeader("Content-Type", "application/x-www-form-urlencoded");
+xhr.onreadystatechange = function() {
+  if (this.readyState === XMLHttpRequest.DONE && this.status === 200) {
+    console.log(xhr.responseText);
+  }
+}
+xhr.send("grant_type=client_credentials&client_id="+clientId+"&client_secret="+clientSecret);
+```
```
sinergise-anze: Well written! May I suggest linking the [official docs](https://docs.sentinel-hub.com/api/latest/reference/#section/Authentication) and maybe using just the `XMLHttpRequest` example? The reason for a single example is that this is not this library's core functionality (it is well documented in SH docs), so it would be best to keep it short.

README.md
```diff
@@ -162,11 +167,79 @@ If we already have a WMS GetMap URL, we can use it directly:
...
+```javascript
+const clientId = /* OAuth Client's id, best to put it in .env file and use it from there */;
+const clientSecret = /* OAuth client's secret, best to put it in .env file and use it from there */;
+
+let xhr = new XMLHttpRequest();
+xhr.open("POST", 'https://services.sentinel-hub.com/oauth/token', true);
+xhr.setRequestHeader("Content-Type", "application/x-www-form-urlencoded");
+xhr.onreadystatechange = function() {
+  if (this.readyState === XMLHttpRequest.DONE && this.status === 200) {
+    console.log(xhr.responseText);
```
sinergise-anze: I would suggest calling `setAuthToken` here instead of calling `console.log`, as this is what most (all?) of the users will do.

.env.example
```diff
@@ -1,5 +1,8 @@
...
+CLIENT_ID=
+CLIENT_SECRET=
+
+# Env var INSTANCE_ID should be set to the ID of some instance in Sentinel Hub
 # Configurator, which contains all the layers listed below it.
-STORYBOOK_INSTANCE_ID=
-STORYBOOK_S2L2A_LAYER_ID=
+INSTANCE_ID=
+S2L2A_LAYER_ID=
+S1GRD_LAYER_ID=
```
sinergise-anze: Are you sure this works? As per [the docs](https://storybook.js.org/docs/configurations/env-vars/), Storybook only imports those env vars that start with `STORYBOOK_` - at least this is how I read them. I agree it is annoying though,as we will either have all vars inappropriately named (they are not Storybook-specific), or we will need to duplicate them.

Alternatively, we could change `package.json` so that all env vars are exported with `STORYBOOK_` prefix, but I'm not sure if this is too much "magic"?

README.md
```diff
@@ -162,11 +167,81 @@ If we already have a WMS GetMap URL, we can use it directly:
...
+})
+  .then(function(response){
+    console.log(response.data.access_token);
+  })
+  .catch(function(error) {
+    console.log(error);
+  });
+```
+
+More about authentication is avilable at [Sentinel Hub documentation](https://docs.sentinel-hub.com/api/latest/#/API/authentication).
```
sinergise-anze: Typo: `avilable`

README.md
```diff
@@ -162,11 +167,81 @@ If we already have a WMS GetMap URL, we can use it directly:
...
+
+- `CLIENT_ID`: OAuth Client's id (optional, authentication is needed for examples that use Processing API)
+- `CLIENT_SECRET`: OAuth Client's secret (optional, authentication is needed for examples that use Processing API)
+
+- `INSTANCE_ID`: id of the configuration instance that will be used in examples
+- `S2L2A_LAYER_ID`: id of the Sentinel-2 L2A layer from that instance
+- `S1GRD_LAYER_ID`: id of the Sentinel-1 GRD layer from that instance
+
+Instance can be created with the [**Configurator**](https://apps.sentinel-hub.com/dashboard/#/configurations) on the **Sentinel Hub Dashboard**.
+It should contain one Sentinel-2 L2A layer and one Sentinel-1 GRD layer .
```
sinergise-anze: Typo: space before the dot.

I would suggest rephrasing this sentence:
```
It should contain at least one Sentinel-2 L2A layer and one Sentinel-1 GRD layer, whose layer IDs should be the same as set in `S2L2A_LAYER_ID` and `S1GRD_LAYER_ID` env vars respectively.
```

example/node/index.js
```diff
@@ -40,43 +61,62 @@ async function run() {
   printOut(`Layer title:`, layer.title);
   printOut('Layer description:', layer.description);
 
-  if (!process.env.AUTH_TOKEN) {
-    printOut('Other examples need an auth token. Set env var AUTH_TOKEN and run again.', null);
+  if (!process.env.CLIENT_ID && !process.env.CLIENT_SECRET) {
```
sinergise-anze: Minor: it should probably be `||` instead of `&&`?

example/node/index.js
```diff
@@ -40,43 +61,62 @@ async function run() {
...
+      'Other examples need an auth token. Set env vars CLIENT_ID and CLIENT_SECRET and run again.',
+      null,
+    );
     return;
   }
 
+  let authToken;
+  const clientId = process.env.CLIENT_ID;
+  const clientSecret = process.env.CLIENT_SECRET;
+  printOut('Requesting auth token with client id and secret from env vars:', { clientId, clientSecret });
```
sinergise-anze: I would suggest _not_ printing out the `clientSecret` - while users might be comfortable with putting it in `.env`, running the examples should not expose it. It is a secret after all. ;)

stories/index.stories.js
```diff
@@ -72,21 +86,50 @@ export const S2GetMap = () => {
...
+      })
+      .catch(function (error) {
+        console.log('Error occurred:', {
+          status: error.response.status,
+          statusText: error.response.status,
+          headers: error.response.headers,
+          data: error.response.data,
+        });
+        return;
+      });
```
sinergise-anze: Since this code is duplicated in many places, may I suggest creating a common function that would fetch auth token from the service (using client id and secret)?

As we discussed, it would be even better if this function was in the library itself, though this is a bit bigger change...

.env.example
zcernigoj: I have no idea why, but it works for me the way it is now. Can you maybe also try?

README.md
zcernigoj: Thanks. Changed and left just the `XMLHttpRequest` example.

README.md
zcernigoj: True, changed.

README.md
zcernigoj: Thanks. Fixed.

README.md
zcernigoj: Thanks. Used your suggestion as it seems better.

example/node/index.js
zcernigoj: Thanks, true. Changed.

example/node/index.js
zcernigoj: I agree. Removed the `clientSecret`.

stories/index.stories.js
zcernigoj: Thanks, true. I moved the code to a common function. I also did that in `example/node/index.js`.

.env.example
sinergise-anze: Interesting, so it does... :)

Sorry for the noise, resolving.

PR #9, opened 2020-01-30
add function to get authToken from OAuth endpoint on sentinelhub

zcernigoj:
Added function to get authToken from OAuth endpoint on sentinelhub (~~`getAuthTokenFromOAuth`~~ `requestAuthToken`). 

Cause: comment on https://github.com/sentinel-hub/sentinelhub-js/pull/8

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,30 +172,20 @@ If we already have a WMS GetMap URL, we can use it directly:
 Requests to Processing API need to be authenticated.
 Documentation about authentication is available at [Sentinel Hub documentation](https://docs.sentinel-hub.com/api/latest/#/API/authentication).
 
-In short, authentication is done by requesting an authentication token and setting it as described in [Layers](#layers).
-The authentication token is retrieved by making a request to the `https://services.sentinel-hub.com/oauth/token` endpoint with the OAuth Client's id and secret.
+In short, authentication is done by getting an authentication token using OAuth Client's id and secret, and setting it as described in [Layers](#layers).
 
 To get the OAuth Client's id and secret, a new OAuth Client must be created in [**User settings**](https://apps.sentinel-hub.com/dashboard/#/account/settings) on **Sentinel Hub Dashboard** under **OAuth clients**.
 OAuth Client's secret is shown only before the creation process is finished so be mindful to save it.
 
-In javascript requesting the authentication token can be done with builtin XMLHttpRequest:
+Getting the authentication token by calling a function with the OAuth Client's id and secret as its parameters:
 
 ```javascript
-const { setAuthToken } = require('sentinelhub-js');
+const { setAuthToken, getAuthTokenFromOAuth } = require('sentinelhub-js');
 
 const clientId = /* OAuth Client's id, best to put it in .env file and use it from there */;
 const clientSecret = /* OAuth client's secret, best to put it in .env file and use it from there */;
-
-let xhr = new XMLHttpRequest();
-xhr.open("POST", 'https://services.sentinel-hub.com/oauth/token', true);
-xhr.setRequestHeader("Content-Type", "application/x-www-form-urlencoded");
-xhr.onreadystatechange = function () {
-  if (this.readyState === XMLHttpRequest.DONE && this.status === 200) {
-    const responseObj = JSON.parse(xhr.response);
-    setAuthToken(xhr.response.access_token);
-  }
-}
-xhr.send("grant_type=client_credentials&client_id=" + clientId + "&client_secret=" + clientSecret);  
+const authToken = await getAuthTokenFromOAuth(clientId, clientSecret);
+setAuthToken(authToken);
 ```
 
 # Examples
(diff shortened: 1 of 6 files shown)


example/node/index.js
```diff
@@ -27,30 +28,12 @@ async function setAuthTokenWithOAuthCredentials() {
     return;
   }
 
-  let authToken;
   const clientId = process.env.CLIENT_ID;
   const clientSecret = process.env.CLIENT_SECRET;
   printOut('Requesting auth token with client id from env vars:', clientId);
 
-  await axios({
```
sinergise-anze: Note that these two lines are no longer needed:
```
const dotenv = require('dotenv').config({ path: '../../.env' });
const axios = require('axios');
```

Similary, import for `axios` in storybook examples is not needed anymore.

stories/legacy.stories.js
```diff
@@ -128,28 +128,9 @@ export const WMSLegacyGetMapFromUrlDatesNotTimes = () => {
 };
 
 async function setAuthTokenWithOAuthCredentials() {
-  let authToken;
```
sinergise-anze: To avoid authenticating every time a new image is fetched (we must set a good example ;) ), I would suggest adding a check here:
```
  if (isAuthTokenSet()) {
    return;
  }
```

EDIT: same in `index.stories.js`.

README.md
```diff
@@ -172,30 +172,20 @@ If we already have a WMS GetMap URL, we can use it directly:
...
 
-In short, authentication is done by requesting an authentication token and setting it as described in [Layers](#layers).
-The authentication token is retrieved by making a request to the `https://services.sentinel-hub.com/oauth/token` endpoint with the OAuth Client's id and secret.
+In short, authentication is done by getting an authentication token using OAuth Client's id and secret, and setting it.
 
 To get the OAuth Client's id and secret, a new OAuth Client must be created in [**User settings**](https://apps.sentinel-hub.com/dashboard/#/account/settings) on **Sentinel Hub Dashboard** under **OAuth clients**.
 OAuth Client's secret is shown only before the creation process is finished so be mindful to save it.
 
-In javascript requesting the authentication token can be done with builtin XMLHttpRequest:
+Getting the authentication token by calling `getAuthTokenFromOAuth()` with the OAuth Client's id and secret as its parameters and then setting the authentication token:
```
sinergise-anze: I know it was my suggestion, but maybe this name is not really perfect (technically, we are getting the token from OAuth _service_)... Not sure I have a better one though - let's discuss live.

stories/legacy.stories.js
zcernigoj: Because of the nature of the `isAuthTokenSet()` function, it returns `true` when the `authToken` is set to `undefined`.  Should we change that so it returns `false` if the `authToken` is `undefined`?

```javascript
export function isAuthTokenSet(): boolean {
  return authToken !== null;
}

File: README.md (hunk 3 above)
Comment (zcernigoj) [reply README.md]: We talked live and agreed to name this function `requestAuthToken` as it tells more about what the function does.

Will rename it.

File: stories/legacy.stories.js (hunk 2 above)
Comment (sinergise-anze) [reply stories/legacy.stories.js]: Yes please! :)

I guess in the future it should also check the auth token expiry time, not top priority yet though (our existing apps handle auth token updating already).

File: example/node/index.js (hunk 1 above)
Comment (zcernigoj) [reply example/node/index.js]: `const dotenv = require('dotenv').config({ path: '../../.env' });` is still needed unfortunately. I moved it above import of sentinelhub-js things.

Removed `const axios = require('axios');` where it was applicable.

File: stories/legacy.stories.js (hunk 2 above)
Comment (zcernigoj) [reply stories/legacy.stories.js]: I changed my mind a little bit - I changed the `setAuthToken()` function so it throws an Error if the parameter is `undefined`. Now we fully comply with the type of `authToken` (either `string` or `null`).

File: README.md (hunk 3 above)
Comment (zcernigoj) [reply README.md]: Did rename it and updated the examples and readme.

File: example/node/index.js (hunk 1 above)
Comment (sinergise-anze) [reply example/node/index.js]: True, I was mislead by the editor that marked variable `dotenv` as unused.

To make it clearee, I would suggest changing the line to just:
```
require('dotenv').config({ path: '../../.env' });
```

But that's already nitpicking, feel free to resolve either way.

File: src/auth.ts [hunk 4]
```diff
@@ -1,13 +1,28 @@
...
 let authToken: string | null = null;
 
 export function getAuthToken(): string | null {
   return authToken;
 }
 
 export function setAuthToken(newAuthToken: string): void {
+  if (newAuthToken === undefined) {
+    throw new Error('Parameter newAuthToken must not be undefined');
+  }
```
sinergise-anze: Maybe a better check would be `if (!newAuthToken)`? Afaict there is nothing special about `undefined`... (since `null` and `""` are also invalid inputs).

zcernigoj: It's a simple one so I changed it as suggested.

example/node/index.js
zcernigoj: It looks better without `const dotenv = ` anyway so I changed it.

PR #10, opened 2020-02-03
Feature/additional datasets

sinergise-anze:
The PR adds support for:
- Sentinel 1 GRD (IW and EW)
- Sentinel 3 (SLSTR and OLCI)


README.md
```diff
@@ -200,11 +200,13 @@ These variables should be put in the `.env` file in the root folder of this proj
...
 - `INSTANCE_ID`: id of the configuration instance that will be used in examples
+- `S1GRDIW_LAYER_ID`: id of the Sentinel-1 GRD IW layer from that instance
+- `S1GRDEW_LAYER_ID`: id of the Sentinel-1 GRD EW layer from that instance
 - `S2L2A_LAYER_ID`: id of the Sentinel-2 L2A layer from that instance
-- `S1GRD_LAYER_ID`: id of the Sentinel-1 GRD layer from that instance
+- ... (see `.env.example` for full list)
 
 Instance can be created with the [**Configurator**](https://apps.sentinel-hub.com/dashboard/#/configurations) on the **Sentinel Hub Dashboard**.
-It should contain at least one Sentinel-2 L2A layer and one Sentinel-1 GRD layer, whose layer IDs should be the same as set in `S2L2A_LAYER_ID` and `S1GRD_LAYER_ID` env vars respectively.
+It should contain at least one Sentinel-2 L2A layer and one Sentinel-1 GRD layer, whose layer IDs should be the same as set in `S2L2A_LAYER_ID` and `S1GRDIW_LAYER_ID` env vars respectively.
```
zcernigoj: I never particularly liked that sentence and now with the storybook examples for S1 GRD (IW & EW) and S3 (OLCI & SLSTR), it's also outdated.

I was thinking we can maybe change it to something similar to "It should contain the layers that are mentioned in the list above."

example/node/index.js
```diff
@@ -56,9 +56,9 @@ async function run() {
     );
     return;
   }
-  if (!process.env.S1GRD_LAYER_ID) {
+  if (!process.env.S1GRDIW_LAYER_ID) {
     printOut(
-      'Example needs id of Sentinel-1 GRD layer to run. Please set env var S1GRD_LAYER_ID and run again',
+      'Example needs id of Sentinel-1 GRD layer to run. Please set env var S1GRDIW_LAYER_ID and run again',
```
zcernigoj: We kinda missed it before, but the Acquisition mode must be set in the configuration, so I suggest we change this to: `Example needs id of Sentinel-1 GRD layer with Acquisition mode set to IW to be able to run.`

README.md
sinergise-anze: Agreed - I fixed it a bit.

example/node/index.js
sinergise-anze: True, but I would suggest we postopne these fixed until after we have more of the datasets ready, because we will need to com up with a consistent way of dealing with more than just S-2 and S-1, if you agree?

zcernigoj: I agree, resolving.

src/layer/LayersFactory.ts
```diff
@@ -9,12 +9,16 @@ import {
   DATASET_MODIS,
   DATASET_AWS_DEM,
   DATASET_AWS_S1GRD_IW,
+  DATASET_S3SLSTR,
+  DATASET_S3OLCI,
```
zcernigoj: Should the `DATASET_AWS_S1GRD_EW` and `S3SLSTRLayer` be imported as well?

README.md
zcernigoj: Nice. I'm nitpicking, now, but maybe change `stories` with `examples`. That way it would cover both node examples and storybook stories.

sinergise-anze: Done. :)

src/layer/LayersFactory.ts
sinergise-anze: As we discussed, I had a branch ready that joined Sentinel 1 IW and EW in a single layer, which I have now merged to this branch. The reason was that apart from `AcquisitionMode`, there are also `Polarization` and `Resolution` parameters which also break dataset into partitions - it would not be consistent to only have `Layer` classes only for `AcquisitionMode`, and having all possible combinations would not be viable.

zcernigoj: Resolving.

src/layer/S1GRDAWSEULayer.ts
```diff
@@ -53,28 +56,41 @@ export class AbstractS1GRDAWSLayer extends AbstractSentinelHubV3Layer {
...
     this.backscatterCoeff = backscatterCoeff;
-    this.resolution = resolution;
   }
 
   protected async updateProcessingGetMapPayload(payload: ProcessingPayload): Promise<ProcessingPayload> {
-    const layerParams = await this.fetchLayerParamsFromSHServiceV3();
+    if (this.polarization === null || this.acquisitionMode === null || this.resolution === null) {
+      if (this.instanceId === null || this.layerId === null) {
+        throw new Error(
+          "Parameters polarization, acquisitionMode and/or resolution are not set, but can't fetch them from service - instanceId and layerId are not available",
```
zcernigoj: Maybe better phrased: 
`Parameters polarization, acquisitionMode and/or resolution are not set and can't be fetched from service because instanceId and layerId are not available.`

sinergise-anze: Much better, thank you!

README.md
zcernigoj: Nice :) Resolving.

src/layer/S1GRDAWSEULayer.ts
zcernigoj: Resolving.

PR #13, opened 2020-02-05
Feature/s 5

sinergise-anze:
This PR adds support for Sentinel-5P dataset. Additionally, it fixes a bug where using WMS API with a specified `evalscript` returned error `400` (evalscript wasn't base64 encoded).


stories/s5pl2.stories.js
```diff
@@ -0,0 +1,320 @@
...
+  img.width = '512';
+  img.height = '512';
+
+  const wrapperEl = document.createElement('div');
+  wrapperEl.innerHTML = "<h2>GetMapUrl (WMS)</h2>";
+  wrapperEl.insertAdjacentElement("beforeend", img);
+
+  const layer = new S5PL2Layer(instanceId, layerId);
+
+  const bbox = new BBox(CRS_EPSG3857, -2035059.4, 15497760.4, -1956787.9, 15576031.8);
```
dgostencnik: Perhaps use some other bbox, I think these are out of s5 reach.

stories/s5pl2.stories.js
```diff
@@ -0,0 +1,320 @@
...
+        return {
+          input: ["CO"],
+          output: { bands: 3 }
+        };
+      }
+
+      function evaluatePixel(sample) {
+        return [2.5 * sample.CO, 2.5 * sample.CO, 2.5 * sample.CO];
+      }
+    `,null, null, null, null, AcquisitionMode.EW, Polarization.DH, Resolution.MEDIUM
```
dgostencnik: These are not valid S5 options (S1 leftovers?)

src/layer/S5PL2Layer.ts
```diff
@@ -0,0 +1,63 @@
...
+  CH4 = 'CH4',
+}
+
+type S5PL2FindTilesDatasetParameters = {
+  type: string;
+  productType: ProductType;
+};
+
+export class S5PL2Layer extends AbstractSentinelHubV3Layer {
+  public readonly dataset = DATASET_S5PL2;
```
dgostencnik: Should `minQa` and `timeliness` also be supported? They are used only by processApi and afaik can't be defined in dasboard for ogc request.

src/layer/wms.ts
```diff
@@ -110,7 +110,7 @@ export function wmsGetMapUrl(
   }
 
   if (evalscript) {
-    queryParams.evalscript = evalscript;
+    queryParams.evalscript = btoa(evalscript);
```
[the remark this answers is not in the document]
sinergise-anze: Good catch! I guess the library should receive it as base64 encoded (in `legacyGetMap*()` functions), then save it unencoded, and encode it back if using WMS. Will fix it, thank you!

src/layer/S5PL2Layer.ts
```diff
@@ -0,0 +1,63 @@
...
+      toTime,
+      maxCount,
+      offset,
+      null,
+      findTilesDatasetParameters,
+    );
+    return {
+      tiles: response.data.tiles.map(tile => {
+        return {
+          geometry: tile.dataGeometry,
```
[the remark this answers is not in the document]
sinergise-anze: Indeed, thank you! Fixed.

src/layer/wms.ts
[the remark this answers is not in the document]
sinergise-anze: Fixed.

stories/s5pl2.stories.js
sinergise-anze: I am not sure about that, but it's probably better to have the same bounding box for all the stories anyway - fixed.

stories/s5pl2.stories.js
sinergise-anze: Oops, thank you, fixed. :)

src/layer/S5PL2Layer.ts
sinergise-anze: I thought about that, but since we won't use them yet anyway, I would suggest leaving them for some future MR, if you agree?

dgostencnik: OK, just create card for this otherwise we'll forget about it as it is not being used yet.

stories/s5pl2.stories.js
dgostencnik: There is still a problem with switched X and Y in AbstractSentinelHubV3Layer.fetchTiles. As a consequence findTiles doesn't return correct results.

dgostencnik: Issue with switched coordinates will be resolved in some other branch.

PR #14, opened 2020-02-06
Fix/various axios interceptor bugs

guss84:
* remove `useCache` and `retries` from `request.config.params` and add `useCache` to `request.config`
* retries are now triggered depending on the error status (5xx)
* remove logging error when retries were exceeded
* add support for setting expiry on cache, and cleanup on library load

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -3,7 +3,7 @@ import axios from 'axios';
 import { GetMapParams, ApiType, PaginatedTiles } from 'src/layer/const';
 import { BBox } from 'src/bbox';
 import { Dataset } from 'src/layer/dataset';
-import { IRequestConfig } from 'src/utils/axiosInterceptors';
+import { RequestConfig } from 'src/utils/axiosInterceptors';
 
 export class AbstractLayer {
   public title: string | null = null;
@@ -20,7 +20,7 @@ export class AbstractLayer {
     switch (api) {
       case ApiType.WMS:
         const url = this.getMapUrl(params, api);
-        const requestConfig: IRequestConfig = { responseType: 'blob', useCache: true };
+        const requestConfig: RequestConfig = { responseType: 'blob', useCache: true };
         const response = await axios.get(url, requestConfig);
         return response.data;
       default:
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -3,6 +3,7 @@ import axios from 'axios';
 import { GetMapParams, ApiType, PaginatedTiles } from 'src/layer/const';
 import { BBox } from 'src/bbox';
 import { Dataset } from 'src/layer/dataset';
+import { IRequestConfig } from 'src/utils/axiosInterceptors';
 
 export class AbstractLayer {
   public title: string | null = null;
@@ -19,7 +20,8 @@ export class AbstractLayer {
     switch (api) {
       case ApiType.WMS:
         const url = this.getMapUrl(params, api);
-        const response = await axios.get(url, { responseType: 'blob' });
+        const requestConfig: IRequestConfig = { responseType: 'blob', useCache: true };
+        const response = await axios.get(url, requestConfig);
         return response.data;
       default:
         const className = this.constructor.name;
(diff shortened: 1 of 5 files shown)


src/utils/axiosInterceptors.ts
```diff
@@ -37,22 +46,25 @@ const fetchCachedResponse = async (request: any): Promise<any> => {
     return request;
   }
 
-  const response = await cache.match(cacheKey);
-  if (response) {
+  const cachedResponse = await cache.match(cacheKey);
+  if (cachedResponse) {
+    if (!hasCachedResponseExpired(cachedResponse)) {
+      return request;
+    }
```
sinergise-anze: Nitpicking: can we negate the check so we can get rid of `if (cachedResponse) {` and lower the nesting level?

Something like this:
```
if (!cachedResponse || !hasCachedResponseExpired(cachedResponse)) {
  return request;
}
...
```

src/utils/axiosInterceptors.ts
```diff
@@ -82,6 +94,14 @@ const setCacheResponse = async (response: any): Promise<any> => {
     }
 
     let cacheKey;
+
+    const expires = new Date();
+    expires.setSeconds(expires.getSeconds() + EXPIRES_IN_SECONDS);
```
sinergise-anze: Cool, except maybe the expiry time should be a parameter when making requests? For some requests it might make sense to expire in a month, and for some (images) half an hour would be better... It depends on the application too I guess.

If you think this is outside of scope, fftr and we can deal with it later.

src/utils/axiosInterceptors.ts
```diff
@@ -82,6 +94,14 @@ const setCacheResponse = async (response: any): Promise<any> => {
...
+    response.headers = {
+      ...response.headers,
+      [EXPIRY_HEADER_KEY]: expires.getTime(),
+    };
```
sinergise-anze: Another thing: I don't think this works? [Date.setSeconds](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setSeconds) expects "an integer between 0 and 59". But even if it does work, it might be best to use UNIX timestamps (`getTime()` and `setTime()`) directly to clearly communicate intent?

```
    const expiresMs = new Date().getTime() + EXPIRES_IN_SECONDS * 1000;
    response.headers = {
      ...response.headers,
      [EXPIRY_HEADER_KEY]: expiresMs,
    };
```

src/layer/AbstractLayer.ts
```diff
@@ -3,6 +3,7 @@ import axios from 'axios';
 import { GetMapParams, ApiType, PaginatedTiles } from 'src/layer/const';
 import { BBox } from 'src/bbox';
 import { Dataset } from 'src/layer/dataset';
+import { IRequestConfig } from 'src/utils/axiosInterceptors';
```
sinergise-anze: Nitpick: since all of the types (which is how we are using `IRequestConfig`, though it's an interface) are named without the leading `I`, can we drop it here too for consistency? (just `RequestConfig`)

guss84: I originally had an issue with this, and decided to leave it as is. I have now realised there were a duplicate in `eslintrc.js`, so the issue is now resolved

src/utils/axiosInterceptors.ts
guss84: Using seconds outside of the "expected" range does work as intended. Further in the docs it is also noted 
> If a parameter you specify is outside of the expected range, setSeconds() attempts to update the date information in the Date object accordingly.

guss84: I have gone with your suggestion though, simply because of intent

src/utils/axiosInterceptors.ts
guss84: Thanks, have gone with what you suggested

src/utils/axiosInterceptors.ts
guss84: I think we can add this in a different PR. So will resolve for now

src/utils/axiosInterceptors.ts
sinergise-anze: Not exactly what I meant, sorry if I wasn't clear enough... `if (cachedResponse) {` still hides a very long block of code. One could avoid this by doing:
```
if (!cachedResponse || !hasCachedResponseExpired(cachedResponse)) {
  return request;
}

request.adapter = async () => {
  ...
};
return request;
```

PR #15, opened 2020-02-07
add findDates()

zcernigoj:
`findDates()` method returns dates for the selected area in the selected time period.

it is done in similar way as `findTiles` for EOCloud layers.


example/node/index.js
```diff
@@ -108,36 +108,57 @@ async function run() {
...
+    getMapParams.bbox,
+    getMapParams.fromTime,
+    getMapParams.toTime,
+    OrbitDirection.ASCENDING,
+  );
+  printOut('dates for S1 GRD', datesS1GRD);
+
+  // // finally, display the image:
+  // const imageUrl = await layerS2L2A.getMapUrl(getMapParams, ApiType.WMS);
+  // printOut('URL of S2 L2A image:', imageUrl);
```
sinergise-anze: This should probably stay uncommented?

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -171,4 +171,43 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       headers: { 'Accept-CRS': 'EPSG:4326' },
     });
   }
+
+  protected fetchDates(
+    bbox: BBox,
+    fromTime: Date,
+    toTime: Date,
+    maxCloudCoverage?: number | null,
```
sinergise-anze: Should we be explicit here that we are using a percentage value? Currently sometimes we are, and sometimes we are not...

Maybe we could:
- rename this to `maxCloudCoveragePercent`, or
- make a special type called `CloudCoveragePercent` and use it here instead of `number`?

src/layer/S1GRDAWSEULayer.ts
```diff
@@ -135,4 +135,22 @@ export class S1GRDAWSEULayer extends AbstractSentinelHubV3Layer {
...
   }
+
+  public async findDates(
+    bbox: BBox,
+    fromTime: Date,
+    toTime: Date,
+    orbitDirection?: OrbitDirection,
+  ): Promise<Date[]> {
+    const findDatesDatasetParameters: S1GRDFindTilesDatasetParameters = {
+      type: this.dataset.shProcessingApiDatasourceAbbreviation,
```
sinergise-anze: Are you certain this is correct for all datasets? Currently, we specify multiple string on the dataset, each of them being used in a different scenario. In this sense, `shProcessingApiDatasourceAbbreviation` is used only when interacting with Processing API. I think we shouldn't use it here, or if we do, we should find a better name for it (so that it is clear it can safely be used with this endpoint).

EDIT: to clarify, I think we should add another identifier to the dataset, which can be used here...

src/layer/S2L1CLayer.ts
```diff
@@ -28,4 +28,14 @@ export class S2L1CLayer extends AbstractSentinelHubV3Layer {
...
+
+  public async findDates(
+    bbox: BBox,
+    fromTime: Date,
+    toTime: Date,
+    maxCloudCoverage?: number,
+  ): Promise<Date[]> {
+    const response = await this.fetchDates(bbox, fromTime, toTime, maxCloudCoverage);
+    return response.data.map(date => new Date(date));
+  }
```
sinergise-anze: Maybe instead of defining `findDates()` in every subclass, we could simply rename `fetchDates()` to `findDates()` (on the parent) and adapt it so that it returns an array of `Date` objects?

stories/s1grdew.stories.js
```diff
@@ -266,6 +266,32 @@ function renderTilesList(containerEl, list) {
...
+
+  const perform = async () => {
+    const data = await layer.findDates(
+      bbox,
+      new Date(Date.UTC(2020, 1 - 1, 1, 0, 0, 0)),
+      new Date(Date.UTC(2020, 1 - 1, 15, 23, 59, 59)),
+      OrbitDirection.ASCENDING,
+    );
+
+    containerEl.innerHTML = "<ul>" + data.map(d => "<li>" + d + "</li>") + "</ul>";
```
sinergise-anze: Nothing major, but this adds a comma (`,`) after each line, which looks a bit weird... this would fix it:
```
  containerEl.innerHTML = "<ul>" + data.map(d => "<li>" + d + "</li>").join("") + "</ul>";
```

src/layer/S2L1CLayer.ts
zcernigoj: I agree, will change.

src/layer/AbstractLayer.ts
```diff
@@ -77,4 +77,13 @@ export class AbstractLayer {
     }
     return flyoverIntervals;
   }
+
+  public findDates(
+    bbox: BBox, // eslint-disable-line @typescript-eslint/no-unused-vars
+    fromTime: any, // eslint-disable-line @typescript-eslint/no-unused-vars
+    toTime: any, // eslint-disable-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
zcernigoj: Thanks, true. Fixed.

example/node/index.js
zcernigoj: Thanks, true. Fixed.

src/layer/S1GRDAWSEULayer.ts
zcernigoj: Thanks, did some testing and found that we were lucky as the `type` is the same as `shProcessingApiDatasourceAbbreviation ` for S1GRD but not necessarily for others.
Also found out that we use `shProcessingApiDatasourceAbbreviation ` in `findTiles` as well (here in `S1GRDAWSEULayer` at least).

Will fix.

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: On the other hand, we are not consistent in the naming elsewhere either, so maybe let's make a separate MR which consolidates "cloud coverage" handling. Resolving.

src/layer/S1GRDAWSEULayer.ts
zcernigoj: Should be fixed now.

src/layer/S2L1CLayer.ts
zcernigoj: Changed. Can you please look if this approach is ok?

stories/s1grdew.stories.js
zcernigoj: Thanks, fixed in all stories.

src/layer/S3OLCILayer.ts
```diff
@@ -23,4 +23,10 @@ export class S3OLCILayer extends AbstractSentinelHubV3Layer {
       hasMore: response.data.hasMore,
     };
   }
+
+  public async findDates(bbox: BBox, fromTime: Date, toTime: Date): Promise<Date[]> {
+    console.log('s3olci layer, finddates', { bbox, fromTime, toTime });
+
+    return super.findDates(bbox, fromTime, toTime);
+  }
```
sinergise-anze: This can probably be removed?

zcernigoj: Thanks, true. Removed.

zcernigoj:
`findDates()` for Landsat 5,7,8 on EO Cloud not yet implemented

src/layer/dataset.ts
```diff
@@ -1,10 +1,27 @@
...
+  S2GM = 'S2GM',
+  S3 = 'S3',
+  S3SLSTR = 'S3SLSTR',
+  S5PL2 = 'S5PL2',
+}
+
+enum EocDatasetParametersType {
+  S1GRD = 'S1GRD',
+  S5PL2 = 'S5PL2',
+}
```
sinergise-anze: I guess that's more of a personal preference, but I find it more readable if strings are used instead of `enum`s in this case. It is also more consistent with how other values are defined in `Dataset`. Do you think we can use the strings directly?

src/layer/S3SLSTRLayer.ts
```diff
@@ -49,4 +49,26 @@ export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
...
+    } = { orbitDirection: OrbitDirection.DESCENDING, view: 'NADIR' },
+  ): Promise<Date[]> {
+    const findDatesDatasetParameters: S3SLSTRFindTilesDatasetParameters = {
+      type: this.dataset.datasetParametersType,
+      orbitDirection: datasetSpecificParameters.orbitDirection,
+      view: datasetSpecificParameters.view,
+    };
+
+    return super.findDates(bbox, fromTime, toTime, {
+      maxCloudCoverage: datasetSpecificParameters && datasetSpecificParameters.maxCloudCoverage,
```
sinergise-anze: Note that if you use this:
```
const x = datasetSpecificParameters && datasetSpecificParameters.maxCloudCoverage;
```

then the value of `x` is either `datasetSpecificParameters.maxCloudCoverage` or `false` - which is probably not we want in this case?

src/layer/S3SLSTRLayer.ts
```diff
@@ -24,7 +24,7 @@ export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
     view: 'NADIR' | 'OBLIQUE' = 'NADIR',
   ): Promise<PaginatedTiles> {
     const findTilesDatasetParameters: S3SLSTRFindTilesDatasetParameters = {
-      type: this.dataset.shProcessingApiDatasourceAbbreviation,
+      type: this.dataset.datasetParametersType,
```
sinergise-anze: Nice, good that you fixed this - can we maybe just name the `datasetParametersType` so that the name explicitly states where it can be used? Doesn't need to be a short name, maybe sth. like `shSearchIndexV3DatasetType`? (not sure if correct, please check before using)

src/layer/S3SLSTRLayer.ts
```diff
@@ -49,4 +49,26 @@ export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
...
+  public async findDates(
+    bbox: BBox,
+    fromTime: Date,
+    toTime: Date,
+    datasetSpecificParameters: {
+      maxCloudCoverage?: number;
+      orbitDirection: OrbitDirection | null;
+      view: 'NADIR' | 'OBLIQUE';
+    } = { orbitDirection: OrbitDirection.DESCENDING, view: 'NADIR' },
+  ): Promise<Date[]> {
```
sinergise-anze: Elsewhere (in `constructor`s, `findTiles`,...) we change the function signature to include the dataset-specific parameters... I find that approach more intuitive, would you agree? In this case, I would suggest changing the function signature to:
```
public async findDates(
    bbox: BBox,
    fromTime: Date,
    toTime: Date,
    maxCloudCoverage: number | null = null;
    orbitDirection: OrbitDirection | null = null;
    view: 'NADIR' | 'OBLIQUE' = 'NADIR';
): Promise<Date[]> {
```

There's also another thing: some of these parameters (like `maxCloudCoverage`) are now a property of a layer, which makes more sense and simplifies the signature of function. But I guess when merging from master, you'll change that anyway (sorry for the extra work... :-/).

src/layer/AbstractLayer.ts
```diff
@@ -79,4 +79,13 @@ export class AbstractLayer {
...
     return flyoverIntervals;
   }
+
+  public async findDates(
+    bbox: BBox, // eslint-disable-line @typescript-eslint/no-unused-vars
+    fromTime: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
+    toTime: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
+    // any additional
+  ): Promise<Date[]> {
+    throw new Error('Not implemented yet');
```
sinergise-anze: I noticed that EOCloud is not supported in this PR yet, do you think we should add it in a separate one?

I don't think the interface will change when we add it, so I don't see a problem, just asking. Fftr.

README.md
```diff
@@ -143,7 +143,7 @@ We can always use layer to search for data availability:
   const layerS2L2A = new S2L2ALayer(instanceId, 'S2L2A', null, null, null, null, null, maxCloudCoverPercent);
   const tilesS2L2A = layerS2L2A.findTiles(bbox, fromDate, toDate, maxCount, offset);
   const flyoverIntervalsS2L2A = layerS2L2A.findFlyoverIntervals(tilesS2L2A.tiles);
-  const dates = layerS2L2A.findDatesUTC(bbox, fromDate, toDate);
+  const datesS2L2A = await layerS2L2A.findDates(bbox, fromTime, toTime, {maxCloudCoverage: cloudCoveragePercent});
```
zcernigoj: We probably want to name the function `findDatesUTC()`

src/layer/AbstractLayer.ts
zcernigoj: Yes, still needs to be added. It can be in this PR I think.

zcernigoj: Support for EOCloud layers was added

src/layer/S2L1CLayer.ts
zcernigoj: `findDates()` for all `Layer`s is now done similar to how `findTiles` is done for EOCloud layers.

src/layer/dataset.ts
zcernigoj: Made a suggested change.

src/layer/S3SLSTRLayer.ts
zcernigoj: Thanks, it was wrong. I think it's not applicable here anymore, but would appreciate a check in the `Abstract...Layer` classes

src/layer/S3SLSTRLayer.ts
zcernigoj: This has moved. But to answer, unfortunately no, can't do such a "personalization" of params which are in all of the dataset objects in `datasets.ts`.

src/layer/S3SLSTRLayer.ts
zcernigoj: I went a different path - I chose the same way the `findTiles` is implemented in EOCloud layers.

zcernigoj:
Still to do:
- tests
- fix return type
- go get additional params if tjey are not set yet.
- check examples and write docs

src/layer/AbstractSentinelHubV1OrV2WithCCLayer.ts
```diff
@@ -34,4 +34,10 @@ export class AbstractSentinelHubV1OrV2WithCCLayer extends AbstractSentinelHubV1O
       cloudCoverPercent: tile.cloudCoverPercentage,
     };
   }
+
+  protected getFindDatesAdditionalParameters(): Record<string, any> {
+    return {
+      maxcc: this.maxCloudCoverPercent / 100,
```
sinergise-anze: I couldn't find a reference that it should be divided by 100, but in my tests there was no difference in results for numbers over `1`, so this code seems to be correct. Feel free to resolve this comment unless you wish to investigate further.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -105,4 +105,31 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+
+    const url = `${this.dataset.findDatesUrl}?${stringify(params, { sort: false })}`;
+    const response = await axios.post(url, payload, {
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept-CRS': 'EPSG:4326',
+      },
+    });
+
+    return response.data.map((d: string) => moment.utc(d));
```
sinergise-anze: In another PR (#41) we are trying to get rid of moment.js objects in the library interface, so it would probably make sense to change this here also.I would suggest something along the lines of:

```
return response.data.map((d: string) => moment.utc(d).toDate());
```

edit: ...in other places too...

src/layer/S3SLSTRLayer.ts
sinergise-anze: I didn't notice it anymore, resolving.

src/layer/S3SLSTRLayer.ts
sinergise-anze: No problem, if anyone searches for this they will see where it is used. We can always rename later if it bothers us, resolving.

stories/landsat8.aws.stories.js
```diff
@@ -163,6 +167,48 @@ export const findTiles = () => {
   return wrapperEl;
 };
 
+export const findDates = () => {
+  const maxCC = 40;
+  const layer = new Landsat8AWSLayer(instanceId, layerId, null, null, null, null, null, maxCC);
+
+  const wrapperEl = document.createElement('div');
+  wrapperEl.innerHTML = `<h2>findTiles for Landsat 8 on AWS with max cloud coverage of ${maxCC}</h2>`;
```
sinergise-anze: Typo: `findTiles` => `findDates`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -105,4 +105,31 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+      timefrom: fromTime.toISOString(),
+      timeto: toTime.toISOString(),
+      ...this.getFindDatesAdditionalParameters(),
+    };
+
+    const url = `${this.dataset.findDatesUrl}?${stringify(params, { sort: false })}`;
+    const response = await axios.post(url, payload, {
+      headers: {
+        'Content-Type': 'application/json',
+        'Accept-CRS': 'EPSG:4326',
```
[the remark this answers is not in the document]
zcernigoj: The storybook examples work without it, so I removed it.

src/layer/AbstractSentinelHubV1OrV2WithCCLayer.ts
zcernigoj: I will leave that for now. We can quickly fix it, if it will present an issue.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
zcernigoj: Thanks, changed. I also changed that the parameters (`fromTime`, `toTime`) of `findDates()` are `Date` objects.

stories/landsat8.aws.stories.js
zcernigoj: Thanks, fixed.

README.md
zcernigoj: Resolving

PR #16, opened 2020-02-07
flyoverIntervals should contain Date objects, not strings

zcernigoj:
FlyoverInterval objects should contain Date objects, not strings


zcernigoj:
This will be fixed in https://github.com/sentinel-hub/sentinelhub-js/pull/17

Closing the Pull Request.

PR #17, opened 2020-02-10
Feature/s1 eocloud

sinergise-anze:
This PR adds support for Sentinel-1 GRD data on EO Cloud (Sentinel Hub v1/2). To make integration easier, I have also refactored `LayersFactory` a bit. 

Note that I have took a bit different approach when implementing `findTiles` compared to `AbstractSentinelHubV3Layer` - subclasses (optionally) implement two translation functions. I find this approach more readable and would suggest using it there too, but would appreciate "independent opinions". :)


src/layer/AbstractLayer.ts
```diff
@@ -32,12 +32,12 @@ export class AbstractLayer {
...
 
-  public findTiles(
+  public async findTiles(
     bbox: BBox, // eslint-disable-line @typescript-eslint/no-unused-vars
-    fromTime: any, // eslint-disable-line @typescript-eslint/no-unused-vars
-    toTime: any, // eslint-disable-line @typescript-eslint/no-unused-vars
-    maxCount?: number, // eslint-disable-line @typescript-eslint/no-unused-vars
-    offset?: number, // eslint-disable-line @typescript-eslint/no-unused-vars
+    fromTime: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
+    toTime: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
```
zcernigoj: if `fromTime` and `toTime` in `Tile` are now `Date`s, the `findFlyoverIntervals()` shoould be changed as well.

Example: `tiles.sort((a, b) => (new Date(a.sensingTime).getTime() > new Date(b.sensingTime).getTime() ? 1 : -1));`, etc.

src/layer/S1GRDEOCloudLayer.ts
```diff
@@ -0,0 +1,105 @@
...
+    throw new Error(
+      `This combination of acquisition mode and polarization (${this.acquisitionMode} / ${
+        this.polarization
+      }) is not available on EO Cloud`,
+    );
+  }
+
+  protected getFindTilesAdditionalParameters(): Record<string, string> {
+    return {
+      productType: 'GRD',
```
zcernigoj: What is the `productType` intended for? Because I can't find it anywhere else in this MR and in other files there is only `type` in `S1GRDAWSEULayer` and `S3SLSTRLayer`.

sinergise-anze: Function `getFindTilesAdditionalParameters()` returns the parameters that are then sent to the backend service and `productType: GRD` is one such parameter...

stories/s1grdiw.eocloud.stories.js
```diff
@@ -0,0 +1,203 @@
...
+        text = JSON.stringify(tile[key]);
+      } else {
+        text = tile[key];
+      }
+      li.innerHTML = `${key} : ${text}`;
+    }
+  });
+}
+
+async function setAuthTokenWithOAuthCredentials () {
```
zcernigoj: This function probably isn't needed in EOCloud stories. If the authentication is needed, it's probably done in a different way than for the S1 on AWS.

src/layer/S1GRDEOCloudLayer.ts
zcernigoj: Ah true. I just overlooked that the `productType` is a parameter for Sentinel-1 on EOCloud.

src/layer/AbstractLayer.ts
sinergise-anze: Thank you, I have removed the Date constructors now.

While at it, I have also fixed sorting function (if equal times were compared, the result depended on the order of arguments).

stories/s1grdiw.eocloud.stories.js
sinergise-anze: Thx, removed.

src/layer/AbstractLayer.ts
zcernigoj: Thanks!
I closed the MR mentioned in the original comment.

Also thanks for fixing the sorting function, I have over-engineered it.

stories/s1grdiw.eocloud.stories.js
zcernigoj: Nice, resolving.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -0,0 +1,128 @@
...
+    const response = await axios.post(url, payload, {
+      headers: { 'Content-Type': 'application/json' },
+    });
+
+    const responseTiles: any[] = response.data.tiles;
+    return {
+      tiles: responseTiles.map(tile => {
+        return {
+          geometry: tile.tileDrawRegionGeometry,
+          sensingTime: new Date(tile.sensingTime),
```
zcernigoj: This way should also be used in other `...Layer`s (`S2L2ALayer`, etc.), because now the `findTiles()` still returns the tiles with `sensingTime` as `string`.

sinergise-anze: True, thanks, fixed.

I have also used ES6 return syntax to make the code a bit more succint.

zcernigoj: Nice, thank you. Resolving

PR #19, opened 2020-02-12
Add Landsat 5, 7 and 8 at EO Cloud datasets and layers

sinergise-anze:
This PR adds support for Landsat 5, 7 and 8 at EO Cloud.

- `cloud coverage` settings are ignored - this will be fixed across all datasets in a future PR
- wrong bbox coordinates order in `AbstractSentinelHubV1OrV2Layer` is fixed ~~(which is also fixed in PR #18) - will rebase / merge once PR #18 is merged to `master`~~ (it was a different bug)

**EDIT:** please ignore TypeScript magic in `makeLayer`, as it is removed in #20 anyway.


src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -76,11 +76,11 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
-          [bbox.maxY, bbox.minX],
-          [bbox.minY, bbox.minX],
-          [bbox.minY, bbox.maxX],
+          [bbox.minX, bbox.minY],
+          [bbox.maxX, bbox.minY],
+          [bbox.maxX, bbox.maxY],
+          [bbox.minX, bbox.maxY],
+          [bbox.minX, bbox.minY],
         ],
       ],
```
zcernigoj: Is there any info about this somewhere to check?

stories/landsat5.eocloud.stories.js
```diff
@@ -0,0 +1,185 @@
...
+
+  const perform = async () => {
+    const layer = new Landsat5EOCloudLayer(
+      instanceId,
+      layerId,
+      `
+        return [2.5 * B04, 1.5 * B03, 0.5 * B02];
+      `,
+    );
+    console.log("aaaaaaaaaaaaaaaaaaa", layer)
```
zcernigoj: :D
Maybe a bit more descriptive message like `Landsat 5 layer with evalscript` would be better, if it's even needed.

stories/landsat5.eocloud.stories.js
```diff
@@ -0,0 +1,185 @@
...
+  wrapperEl.insertAdjacentElement("beforeend", containerEl);
+
+  const perform = async () => {
+    const data = await layer.findTiles(
+      bbox,
+      new Date(Date.UTC(2000, 1 - 1, 1, 0, 0, 0)),
+      new Date(Date.UTC(2020, 1 - 1, 15, 23, 59, 59)),
+      5,
+      null,
+      OrbitDirection.ASCENDING,
```
zcernigoj: This probably doesn't belong here.

stories/landsat7.eocloud.stories.js
```diff
@@ -0,0 +1,184 @@
...
+  wrapperEl.insertAdjacentElement("beforeend", containerEl);
+
+  const perform = async () => {
+    const data = await layer.findTiles(
+      bbox,
+      new Date(Date.UTC(1990, 1 - 1, 1, 0, 0, 0)),
+      new Date(Date.UTC(2020, 1 - 1, 15, 23, 59, 59)),
+      5,
+      null,
+      OrbitDirection.ASCENDING,
```
zcernigoj: This probably doesn't belong here.

stories/landsat8.eocloud.stories.js
```diff
@@ -0,0 +1,184 @@
...
+  wrapperEl.insertAdjacentElement("beforeend", containerEl);
+
+  const perform = async () => {
+    const data = await layer.findTiles(
+      bbox,
+      new Date(Date.UTC(2000, 1 - 1, 1, 0, 0, 0)),
+      new Date(Date.UTC(2020, 1 - 1, 15, 23, 59, 59)),
+      5,
+      null,
+      OrbitDirection.ASCENDING,
```
zcernigoj: This probably doesn't belong here.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
sinergise-anze: This SO answer seems to summarize it: https://stackoverflow.com/a/57653830 - or the linked [GeoJSON specs](https://tools.ietf.org/html/rfc7946#section-3.1.1):

> The first two elements are longitude and latitude, or
   easting and northing, precisely in that order and using decimal
   numbers.

That said, we are inconsistent in that we use x and y instead of longitude/latitude or easting/northing... Still, I think this fix makes it _less wrong_, but we should revisit how we handle CRSs and BBoxes. I would however prefer to leave this for the future if you agree?

stories/landsat5.eocloud.stories.js
sinergise-anze: :D Thanks, fixed. I have already noticed and fixed it in #20, but I agree, it belongs here, so I have backported the change to this branch.

stories/landsat5.eocloud.stories.js
sinergise-anze: Thx, fixed.

stories/landsat7.eocloud.stories.js
sinergise-anze: Thx, fixed.

stories/landsat8.eocloud.stories.js
sinergise-anze: Thx, fixed.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
zcernigoj: Thanks!
Yeah, of course we can revisit this in the future. I just wanted to know what it is referred to, so the links are more than enough.

stories/landsat5.eocloud.stories.js
zcernigoj: Resolving.

stories/landsat5.eocloud.stories.js
zcernigoj: Resolving.

stories/landsat7.eocloud.stories.js
zcernigoj: Resolving.

stories/landsat8.eocloud.stories.js
zcernigoj: Resolving.

PR #20, opened 2020-02-13
Fix/cloud cover percent

sinergise-anze:
Note: this PR is against #19's branch.

This PR makes use of various parameters (like `maxCloudCoverPercent`) more consistent across the datasets:
- all the relevant parameters are specified when instantiating the layer
- `findTiles()` has the same parameters across all datasets (though the layer itself will supply others as needed)
- all cloud cover (internal) variables are now named consistently: `cloudCoverPercent` (if value) or `maxCloudCoverPercent` (if limit)

Also:
- makeLayer() was moved from `AbstractSentinelHubV1OrV2WithCCLayer` directly to `LandsatXEOCloudLayer` (X ∈  {5, 7, 8}) because I couldn't make TypeScript "magic" work so that it would create appropriate child class instance

### DIFF
--- a/src/layer/S1GRDEOCloudLayer.ts
+++ b/src/layer/S1GRDEOCloudLayer.ts
@@ -2,7 +2,7 @@ import { BackscatterCoeff } from 'src/layer/const';
 import { DATASET_EOCLOUD_S1GRD } from 'src/layer/dataset';
 
 import { AbstractSentinelHubV1OrV2Layer } from 'src/layer/AbstractSentinelHubV1OrV2Layer';
-import { AcquisitionMode, Polarization, Resolution } from 'src/layer/S1GRDAWSEULayer';
+import { AcquisitionMode, Polarization, Resolution, OrbitDirection } from 'src/layer/S1GRDAWSEULayer';
 
 /*
   Note: the usual combinations are IW + DV/SV + HIGH and EW + DH/SH + MEDIUM.
@@ -16,6 +16,7 @@ export class S1GRDEOCloudLayer extends AbstractSentinelHubV1OrV2Layer {
   protected orthorectify: boolean | null = false;
   protected backscatterCoeff: BackscatterCoeff | null = BackscatterCoeff.GAMMA0_ELLIPSOID;
   protected resolution: Resolution | null = null;
+  protected orbitDirection: OrbitDirection | null = null;
 
   public constructor(
     instanceId: string | null,
@@ -26,6 +27,7 @@ export class S1GRDEOCloudLayer extends AbstractSentinelHubV1OrV2Layer {
     description: string | null = null,
     acquisitionMode: AcquisitionMode | null = null,
     polarization: Polarization | null = null,
+    orbitDirection: OrbitDirection | null = null,
   ) {
     super(instanceId, layerId, evalscript, evalscriptUrl, title, description);
     // it is not possible to determine these parameters by querying the service, because there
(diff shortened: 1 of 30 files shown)


src/layer/S1GRDEOCloudLayer.ts
```diff
@@ -95,11 +98,16 @@ export class S1GRDEOCloudLayer extends AbstractSentinelHubV1OrV2Layer {
...
+  protected getFindTilesAdditionalParameters(): Record<string, any> {
+    const result = {
       productType: 'GRD',
       acquisitionMode: this.acquisitionMode,
       polarization: this.polarization,
+      orbitDirection: this.orbitDirection,
     };
+    if (this.orbitDirection !== null) {
+      result.orbitDirection = this.orbitDirection;
+    }
```
zcernigoj: One of those is unnecessary, either put the `orbitDirection` in the `result` without checking or put it when checking but not before.

src/layer/S3SLSTRLayer.ts
```diff
@@ -12,29 +13,58 @@ type S3SLSTRFindTilesDatasetParameters = {
...
+  }
+
+  protected async updateProcessingGetMapPayload(payload: ProcessingPayload): Promise<ProcessingPayload> {
+    payload.input.data[0].dataFilter.maxCloudCoverage = this.maxCloudCoverPercent;
+    return payload;
+  }
+
+  protected getWmsGetMapUrlAdditionalParameters(): Record<string, any> {
+    return {
+      maxcc: this.maxCloudCoverPercent,
```
zcernigoj: Would `orbitDirection` and `view` also be needed here?

src/legacyCompat.ts
```diff
@@ -145,9 +145,12 @@ function parseLegacyWmsGetMapParams(wmsParams: Record<string, any>): ParsedLegac
...
   }
 
-  if (params.maxcc) {
-    getMapParams.maxCCPercent = parseInt(params.maxcc);
-  }
+  // Instead of dealing with maxCC at this point, we pass it through as an unknown parameter. In the
+  // future, it should be used to instantiate the layers instead.
+  // if (params.maxcc) {
+  //   getMapParams.maxCloudCoverPercent = parseInt(params.maxcc);
+  // }
```
zcernigoj: This could probably be deleted right before merging when it's all ready.

src/layer/S1GRDEOCloudLayer.ts
sinergise-anze: Thank you, fixed.

src/legacyCompat.ts
sinergise-anze: Actually I left it here on purpose, as a message - it is a reminder that this is not done properly, because `maxcc` is something that we _do_ know how to handle, so passing it through `unknown` params is not correct. Still, without the named parameters in constructors this wolud be difficult to do - so I would like to leave it for later (when the constructors are changed) if you agree?

src/layer/S3SLSTRLayer.ts
sinergise-anze: Maybe, not sure to be honest... I can't find them as WMS parameters so I'm no sure if they are supported. 

As discussed, we should probably leave this for now, and will revisit this question when we make the test suite for S-3.

zcernigoj: Agreed. Resolving.

src/layer/S1GRDEOCloudLayer.ts
zcernigoj: Nice. Resolving.

src/legacyCompat.ts
zcernigoj: Yeah of course, let's leave it. Resolving.

PR #21, opened 2020-02-14
Feature/tests

sinergise-anze:
This PR configures `jest`:

- `expect` is extended with methods `toHaveOrigin`, `toHaveQueryParamsValues` and `toHaveQueryParams`
- tests are run automatically through GitHub Actions
- a small test is written for `WmsLayer` (on a non-SentinelHub host, to avoid SH-specific parameters)
- along the way, two small bugs are fixed in `WmsLayer`:
  - parameters `transparent` and `showlogo` are removed
  - trailing ampersand (`&`) is removed

Note that mocking requests is not part of this PR - the test is done on `getMapUrl()` method which simply returns a string.


jest-setup.ts
```diff
@@ -0,0 +1,77 @@
...
+    return {
+      message: () => `URL hostname should not have value [${expectedOrigin}], but it does`, // if .not is used
+      pass: true,
+    };
+  },
+});
+
+/* ************************ */
+
+function breakUrl(urlWithQueryParams: string) {
```
dgostencnik: Perhapse parseUrl would be better name for this function.

package.json
```diff
@@ -31,13 +34,24 @@
...
     "build-storybook": "build-storybook"
   },
+  "jest": {
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
+      "js"
+    ],
+    "transform": {
+      "\\.(ts|tsx)$": "<rootDir>/node_modules/ts-jest/preprocessor.js"
```
dgostencnik: jest doesn't like preprocessor (anymore)
```
(WARN) Replace any occurrences of "ts-jest/dist/preprocessor.js" or  "<rootDir>/node_modules/ts-jest/preprocessor.js" 
in the 'transform' section of your Jest config with just "ts-jest".
```

jest-setup.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (params[k] === undefined) {
+        return {
+          message: () => `URL query parameter [${k}] should exist, but it doesn't`,
+          pass: false,
+        };
+      }
+    }
+    return {
+      message: () =>
+        `URL [${received}] should not include all of the parameters ${JSON.stringify(expectedParamsKeys)}, but it does`,
```
dgostencnik: This is a bit confusing, but if I understand correctly this is used when with .not ?

src/layer/__tests__/WmsLayer.ts
```diff
@@ -0,0 +1,38 @@
...
+    bbox: bbox,
+    fromTime: new Date(Date.UTC(2020, 1 - 1, 10, 0, 0, 0)), // 2020-01-10/2020-01-10
+    toTime: new Date(Date.UTC(2020, 1 - 1, 10, 23, 59, 59)),
+    width: 512,
+    height: 512,
+    format: MimeTypes.JPEG,
+  };
+  const imageUrl = layer.getMapUrl(getMapParams, ApiType.WMS);
+
+  expect(imageUrl).toHaveOrigin('https://proba-v-mep.esa.int');
```
dgostencnik: Should baseUrl also be tested somewhere?

jest-setup.ts
sinergise-anze: Yes, I agree it is a bit confusing - and your understanding is correct (or at least the same as mine ;) ). However this is not something I can do much about...

package.json
sinergise-anze: Thank you, missed that! I have now also added `esModuleInterop` config option to suppress another warning, and fixed Axios interceptors (they expected `caches` to be defined which is not the case when running tests).

jest-setup.ts
sinergise-anze: Yes, much better, thank you.

src/layer/__tests__/WmsLayer.ts
sinergise-anze: Maybe, though this PR and #22 are mostly about setting up the tools that we will need for testing, not so much about the tests themselves. I did add a test just to show that these tools work, but I would prefer to keep them small and we can complete them in the future, if you agree?

dgostencnik: OK

PR #22, opened 2020-02-14
Feature/tests mock

sinergise-anze:
This PR depends on #21 - it adds the ability to check if `axios.get()` was called with correct parameters.

### DIFF
--- a/src/layer/__tests__/WmsLayer.ts
+++ b/src/layer/__tests__/WmsLayer.ts
@@ -56,7 +56,7 @@ test('WmsLayer.getMap makes an appropriate request', () => {
   };
   layer.getMap(getMapParams, ApiType.WMS);
 
-  expect(axios.get.mock.calls.length).toBe(1);
+  expect(axios.get).toHaveBeenCalledTimes(1);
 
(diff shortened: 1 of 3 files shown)


src/layer/__tests__/WmsLayer.ts
```diff
@@ -36,3 +37,45 @@ test('WmsLayer.getMapUrl returns an URL', () => {
...
+    bbox: bbox,
+    fromTime: new Date(Date.UTC(2020, 1 - 1, 10, 0, 0, 0)), // 2020-01-10/2020-01-10
+    toTime: new Date(Date.UTC(2020, 1 - 1, 10, 23, 59, 59)),
+    width: 512,
+    height: 512,
+    format: MimeTypes.JPEG,
+  };
+  layer.getMap(getMapParams, ApiType.WMS);
+
+  expect(axios.get.mock.calls.length).toBe(1);
```
dgostencnik: this one could be replaced with
```
expect(axios.get).toHaveBeenCalledTimes(1);
```

sinergise-anze: It does look a bit easier to understand, so I have replaced it, thx.

PR #26, opened 2020-02-21
Fix imports in the README examples

sinergise-anze:
The import statements in the README were wrong, this PR fixes them.


zcernigoj:
Not sure about the line 186: `const { setAuthToken, requestAuthToken } = require('sentinelhub-js');`

sinergise-anze:
> Line 186 also needs to be changed to `... require('@sentinel-hub/sentinelhub-js');`

Thanks, good catch! Fixed.

PR #27, opened 2020-02-21
Feature/public layer properties

sinergise-anze:
EO Browser needs to have access to layer properties (such as `acquisitionMode`, `polarization`,... for Sentinel 1) for filtering the layers.

### DIFF
--- a/src/layer/S3SLSTRLayer.ts
+++ b/src/layer/S3SLSTRLayer.ts
@@ -13,7 +13,7 @@ type S3SLSTRFindTilesDatasetParameters = {
 
 export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
   public readonly dataset = DATASET_S3SLSTR;
-  protected maxCloudCoverPercent: number;
+  public maxCloudCoverPercent: number;
   protected orbitDirection: OrbitDirection | null;
   protected view: 'NADIR' | 'OBLIQUE';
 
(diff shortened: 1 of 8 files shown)


src/layer/S5PL2Layer.ts
```diff
@@ -30,7 +30,7 @@ type S5PL2FindTilesDatasetParameters = {
 export class S5PL2Layer extends AbstractSentinelHubV3Layer {
   public readonly dataset = DATASET_S5PL2;
   protected productType: ProductType;
-  protected maxCloudCoverPercent: number;
+  public maxCloudCoverPercent: number;
```
zcernigoj: Should the `minQa` and `productType` also be public?

src/layer/S3SLSTRLayer.ts
```diff
@@ -13,7 +13,7 @@ type S3SLSTRFindTilesDatasetParameters = {
 
 export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
   public readonly dataset = DATASET_S3SLSTR;
-  protected maxCloudCoverPercent: number;
+  public maxCloudCoverPercent: number;
```
zcernigoj: Should the `orbitDirection` and `view` also be public?

src/layer/S5PL2Layer.ts
zcernigoj: Already done, resolving.

src/layer/S3SLSTRLayer.ts
zcernigoj: Already done, resolving.

sinergise-anze: Thanks, fixed both here and in S-5.

PR #29, opened 2020-02-24
Feature/find flyovers coverage

sinergise-anze:
This PR implements calculation of coverage in `findFlyovers()` (previously known as `findFlyoverIntervals()`). I had reservations about whether this should be implemented inside this function or outside (for performance reasons), but it would complicate interface a bit, so I decided to add it in.

`@turf/union` didn't have TypeScript types definitions ready, so I had to come up with my own. Not an expert here, would appreciate feedback.

**EDIT**: The interface for `findFlyovers()` is now changed so that it fetches tiles itself, thus avoiding possible user mistakes (ignoring `hasMore`).

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -9,6 +9,7 @@ import area from '@turf/area';
 import union from '@turf/union'; // @turf/union is missing types definitions, we supply them separately
 
 import { Polygon, MultiPolygon } from '@turf/helpers';
+import { CRS_EPSG4326 } from 'src/crs';
 
 export class AbstractLayer {
   public title: string | null = null;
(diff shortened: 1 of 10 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -44,41 +50,104 @@ export class AbstractLayer {
...
+      flyoverIntervals[flyoverIndex].coveragePercent = this.calculateCoveragePercent(
+        bbox,
+        currentFlyoverGeometry,
+      );
+    }
     return flyoverIntervals;
   }
 
+  private calculateCoveragePercent(bbox: BBox, flyoverGeometry: Polygon | MultiPolygon): number {
+    const bboxGeometry: Polygon = {
```
guss84: Feel free to resolve this and merge, but maybe we should try and add function to convert a bbox to a geometry as this is duplicated in a few spots already:
https://github.com/sentinel-hub/sentinelhub-js/blob/master/src/layer/AbstractSentinelHubV3Layer.ts#L163
https://github.com/sentinel-hub/sentinelhub-js/blob/master/src/layer/AbstractSentinelHubV1OrV2Layer.ts#L79

sinergise-anze: Sounds like a good idea, yes. I would prefer todo it in a separate PR though, since this one is kind of needed asap... :)

PR #30, opened 2020-02-25
Cleanup: bbox should know how to convert itself to GeoJSON to avoid code duplication

sinergise-anze:
Code cleanup.

`BBox` instance `toGeoJSON()`'s result had to be defined as `:any`, because turf's `Polygon` doesn't allow `crs` to be specified, and we need it when specifying payload for Sentinel Hub Processing API.

### DIFF
--- a/src/bbox.ts
+++ b/src/bbox.ts
@@ -20,4 +20,22 @@ export class BBox {
     this.minY = minY;
     this.maxY = maxY;
   }
+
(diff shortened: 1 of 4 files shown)


sinergise-anze:
> Different order of `maxY`, `minY` in the files before the change. Needs to be checked if the new way applies for all:
> 
> ...

As far as I can see, the only difference is in where the rectangle starts and which direction is used to draw it... which shouldn't matter. Am I missing something?

PR #31, opened 2020-02-25
Return meta.averageCloudCoverPercent from findFlyovers()

sinergise-anze:
In some cases it is beneficial to know the cloud cover percentage for flyovers (for datasets that support it).

Note that "average" is calculated very imprecisely - it is just the average value of the CC of the tiles that intersect the bounding box, no matter how big the intersection is. This can be improved in the future if needed.


zcernigoj:
I will merge this and publish to npm with the https://github.com/sentinel-hub/sentinelhub-js/pull/32

PR #34, opened 2020-02-27
Fix findTiles() for Landsat8 on AWS

sinergise-anze:
- implements  default `findTiles()` on all V3 layers (fixes DEM, MODIS)
- introduces `AbstractSentinelHubV3WithCCLayer` layer to implement CC-specific functionality there (`findTiles` and constructor moved from both of the S-2 layers)
- creates storybook for Landsat 8 (AWS)


src/layer/AbstractSentinelHubV3WithCCLayer.ts
```diff
@@ -0,0 +1,49 @@
+import { BBox } from 'src/bbox';
+import { PaginatedTiles } from 'src/layer/const';
+import { AbstractSentinelHubV3Layer } from './AbstractSentinelHubV3Layer';
+
+// this class provides any SHv3-specific functionality to the subclasses:
```
zcernigoj: Maybe change the comment a little so it's not exactly the same as in `src/layer/AbstractSentinelHubV3Layer.ts`? Maybe in the same fashion as in `AbstractSentinelHubV1OrV2WithCCLayer.ts`.
For example: `// same as AbstractSentinelHubV3Layer, but with maxCloudCoverPercent (useful for Landsat datasets)`

zcernigoj: Ok, I changed it to `// same as AbstractSentinelHubV3Layer, but with maxCloudCoverPercent (for layers which support it)`.

Resolving

PR #36, opened 2020-02-28
Fix/find flyovers errors

sinergise-anze:
- uses [polygon-clipping](https://github.com/mfogel/polygon-clipping) instead of Turf union/intersect due to numerous bugs (invalid intersection calculation, sometimes hangs)
- the tiles are now fetched on-the-go, as we are calculating the flyovers (less memory consumption)
- improved robustness (better error handling)

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -1,14 +1,12 @@
 import axios from 'axios';
 
-import { GetMapParams, ApiType, Tile, PaginatedTiles, FlyoverInterval } from 'src/layer/const';
+import { GetMapParams, ApiType, PaginatedTiles, FlyoverInterval } from 'src/layer/const';
 import { BBox } from 'src/bbox';
 import { Dataset } from 'src/layer/dataset';
 import { RequestConfig } from 'src/utils/axiosInterceptors';
-import intersect from '@turf/intersect';
 import area from '@turf/area';
-import union from '@turf/union'; // @turf/union is missing types definitions, we supply them separately
+import { union, intersection, Geom } from 'polygon-clipping';
 
-import { Polygon, MultiPolygon, Feature } from '@turf/helpers';
 import { CRS_EPSG4326 } from 'src/crs';
 
 export class AbstractLayer {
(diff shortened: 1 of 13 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -60,21 +58,103 @@ export class AbstractLayer {
...
+
+    let flyovers: FlyoverInterval[] = [];
+    let flyoverIndex = 0;
+    let currentFlyoverGeometry: Geom = null;
+    let nTilesInFlyover;
+    let sumCloudCoverPercent;
     for (let i = 0; i < maxFindTilesRequests; i++) {
-      const { tiles: partialTiles, hasMore } = await this.findTiles(
+      // grab new batch of tiles:
+      const { tiles, hasMore } = await this.findTiles(
```
dgostencnik: This algorithm  assumes tiles will always be sorted descending ( from most to least recent). Perhaps some parameter could be added to findTiles to make this more explicit. No need to do it in scope of this MR.

src/layer/AbstractLayer.ts
```diff
@@ -60,21 +58,103 @@ export class AbstractLayer {
...
+              bboxGeometry,
+              bboxArea,
+              currentFlyoverGeometry,
+            );
+          } catch (err) {
+            // if anything goes wrong, play it safe and set coveragePercent to null:
+            flyovers[flyoverIndex].coveragePercent = null;
+          }
+          if (sumCloudCoverPercent !== undefined) {
+            flyovers[flyoverIndex].meta.averageCloudCoverPercent = sumCloudCoverPercent / nTilesInFlyover;
```
dgostencnik: As discussed in person, `coveragePercent` should also be taken into account, but I think for now this is good enough.

stories/s1grdiw.eocloud.stories.js
```diff
@@ -214,6 +214,49 @@ export const findTilesEPSG4326 = () => {
...
+  const img = document.createElement('img');
+  img.width = '512';
+  img.height = '512';
+  wrapperEl.insertAdjacentElement("beforeend", img);
+
+  const flyoversContainerEl = document.createElement('pre');
+  wrapperEl.insertAdjacentElement("beforeend", flyoversContainerEl);
+
+  const perform = async () => {
+    const layer = (await LayersFactory.makeLayers(`${DATASET_EOCLOUD_S1GRD.shServiceHostname}v1/wms/${instanceId}`, (lId, datasetId) => (layerId === lId)))[0];
```
dgostencnik: That doesn't work. Probably wrong layer name, other examples in this story also fails with 500 because of nonexistent layer `S1`

sinergise-anze: The `.env` example in internal documentation had the wrong value, which caused this example not to work (I have already fixed it). I think if you update `.env` it should work.

src/layer/AbstractLayer.ts
sinergise-anze: Yes, agree completely, but would prefer to solve this in a future MR.

PR #41, opened 2020-03-09
Fix: use Date objects in library interface

sinergise-anze:
This PR changes interface to use JS `Date` objects again in `findTiles` and `findFlyovers` instead of moment.js objects (which was a temporary measure introduced in https://github.com/sentinel-hub/sentinelhub-js/pull/38 to ensure correctness). 

Original idea was to use day.js library instead of moment.js, but after testing it turns out that our apps are bigger with day.js (because they already include moment.js). 

The test checks that date handling is correct by mocking response from the server, and checking that the returned `Date` object is correct. Note that we now test against node.js 8, 10 and 12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,17 +141,17 @@ We can always use layer to search for data availability:
 
   const maxCloudCoverPercent = 50;
   const layerS2L2A = new S2L2ALayer(instanceId, 'S2L2A', null, null, null, null, null, maxCloudCoverPercent);
-  const { tiles, hasMore } = layerS2L2A.findTiles(bbox, fromDate, toDate, maxCount, offset);
-  const flyoversS2L2A = layerS2L2A.findFlyovers(bbox, fromDate, toDate);
-  const dates = layerS2L2A.findDatesUTC(bbox, fromDate, toDate);
+  const { tiles, hasMore } = layerS2L2A.findTiles(bbox, fromTime, toTime, maxCount, offset);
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -141,17 +141,17 @@ We can always use layer to search for data availability:
 
   const maxCloudCoverPercent = 50;
   const layerS2L2A = new S2L2ALayer(instanceId, 'S2L2A', null, null, null, null, null, maxCloudCoverPercent);
-  const { tiles, hasMore } = layerS2L2A.findTiles(bbox, fromDate, toDate, maxCount, offset);
-  const flyoversS2L2A = layerS2L2A.findFlyovers(bbox, fromDate, toDate);
-  const dates = layerS2L2A.findDatesUTC(bbox, fromDate, toDate);
+  const { tiles, hasMore } = layerS2L2A.findTiles(bbox, fromTime, toTime, maxCount, offset);
```
zcernigoj: Should we be more transparent and also define `fromTime` and `toTime`? 
One reason being that we now decided that it should be the `Date` object.
And another is that `bbox` is defined in the `Fetching images` part of readme, but for  `fromTime` and `toTime` you can only conclude they are `Date()` objects.

sinergise-anze: Hm, not sure... On one hand, it would be nice, but they can already see that we are using `Date` objects from before (example for `getMapParams`). I would leave this for now, if you agree, and tackle it along with the interface change (named parameters).

PR #44, opened 2020-03-11
Specify orbitDirection with S-3 SLSTR to avoid blank images

sinergise-anze:
The images with S-3 SLSTR appear empty when orbit direction is ASC, so we need to specify this parameter when using Processing API.

### DIFF
--- a/src/layer/S3SLSTRLayer.ts
+++ b/src/layer/S3SLSTRLayer.ts
@@ -32,13 +32,14 @@ export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
   ) {
     super(instanceId, layerId, evalscript, evalscriptUrl, dataProduct, title, description);
     this.maxCloudCoverPercent = maxCloudCoverPercent;
-    // images that are not DESCENDING are blank, so we can hardcode this:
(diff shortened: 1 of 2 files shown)


src/layer/S3SLSTRLayer.ts
```diff
@@ -32,13 +32,14 @@ export class S3SLSTRLayer extends AbstractSentinelHubV3Layer {
...
     this.maxCloudCoverPercent = maxCloudCoverPercent;
-    // images that are not DESCENDING are blank, so we can hardcode this:
+    // images that are not DESCENDING appear empty:
     this.orbitDirection = OrbitDirection.DESCENDING;
     this.view = view;
   }
 
   protected async updateProcessingGetMapPayload(payload: ProcessingPayload): Promise<ProcessingPayload> {
     payload.input.data[0].dataFilter.maxCloudCoverage = this.maxCloudCoverPercent;
+    payload.input.data[0].dataFilter.orbitDirection = this.orbitDirection;
```
Ardweaden: Not really a comment regarding the code directly, but I would assume `view` would have to be added to payload at some point as well? Right now we search by `NADIR`, which is also the default for Processing API.

PR #46, opened 2020-03-12
Fix/named parameters

sinergise-anze:
This PR changes the way layers are instantiated - instead of normal paramaters, "named" parameters are used (JS doesn't have support for real named parameters, so they are really objects). 

I wanted to make sure that the users will be able to find the list of available parameters for each of the classes, so I have also installed `typedoc`. Please run `npm run build-doc` and explore resulting `doc/index.html` - I would appreciate some feedback.

Apart from that, a small bug was fixed along the way (S-3 SLSTR view was not being passed to Processing API). 

Two of the stories did not work (wrong dates) so I added a notice about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,36 +35,57 @@ $ npm install @sentinel-hub/sentinelhub-js
 
 ## Layers
 
-The core data structure is `Layer`, which corresponds to a _layer_ as returned by OGC WMS GetCapabilities request. If the URL matches one of the Sentinel Hub service URLs, additional capabilities are unlocked.
-
-Basic (WMS-capable) `Layer` can be initialized like this:
+The core data structure is `Layer`, which corresponds to a _layer_ as returned by OGC WMS GetCapabilities request. Basic (WMS-capable) `Layer` can be initialized like this:
 
 ```javascript
   import { WmsLayer } from '@sentinel-hub/sentinelhub-js';
 
-  const layer = new WmsLayer('https://services.sentinel-hub.com/ogc/wms/<your-instance-id>', '<layer-id>', 'Title', 'Description');
+  const layer = new WmsLayer({
+    url: 'https://services.sentinel-hub.com/ogc/wms/<your-instance-id>',
+    layerId: '<layer-id>',
+  });
 ```
 
-Such layer would only allow WMS requests. However, `Layer` is also a superclass for multiple dataset-specific subclasses (like `S1GRDAWSEULayer` - Sentinel-1 GRD data on AWS eu-central-1 Sentinel Hub endpoint) which can be instantiated with their own specific parameters and thus unlock additional powers.
+Such layer would only allow WMS requests. However, `Layer` is also a superclass for multiple dataset-specific subclasses (like `S1GRDAWSEULayer` - Sentinel-1 GRD data on AWS eu-central-1 Sentinel Hub endpoint) which can be instantiated with their own specific parameters and thus have additional capabilities.
 
 When it comes to Sentinel Hub layers, there are four ways to determine their content:
 
-- by `layerId`: ID of the layer, as used by OGC WMS and SentinelHub Configurator
+- by `instanceId` and `layerId`: ID of the layer, as used by OGC WMS and SentinelHub Configurator
 - by `evalscript`: custom (javascript) code that will be executed by the service per each pixel and will calculate the (usually RGB/RGBA) values
 - by `evalscriptUrl`: the URL from which the evalscript can be downloaded from
-- by `dataProduct`: the structure which contains an ID of a pre-existing product
+- by `dataProduct`: the ID of a pre-existing data product
 
 ```javascript
   import { S1GRDAWSEULayer } from '@sentinel-hub/sentinelhub-js';
 
   let layerS1;
-  layerS1 = new S1GRDAWSEULayer(instanceId, '<layer-id>', null, null, null, 'Title', 'Description');
-  layerS1 = new S1GRDAWSEULayer(null, null, myEvalscript, null, null, 'Title', 'Description', 'IW', 'DV', 'HIGH');
-  layerS1 = new S1GRDAWSEULayer(null, null, null, myEvalscriptUrl, null, 'Title', 'Description', 'IW', 'DV', 'HIGH');
-  layerS1 = new S1GRDAWSEULayer(null, null, null, null, '<data-product-id>', 'Title', 'Description', 'EW', 'DH', 'MEDIUM');
+  layerS1 = new S1GRDAWSEULayer({
+    instanceId: '<my-instance-id>',
+    layerId: '<layer-id>',
+  );
+  layerS1 = new S1GRDAWSEULayer({
+    evalscript: myEvalscript,
+    title: 'Title',
+    description: 'Description',
+    acquisitionMode: AcquisitionMode.IW,
+    polarization: Polarization.DV,
+    resolution: Resolution.HIGH,
+  });
+  layerS1 = new S1GRDAWSEULayer({
+    evalscriptUrl: myEvalscriptUrl,
+    acquisitionMode: AcquisitionMode.IW,
+    polarization: Polarization.DV,
+    resolution: Resolution.HIGH,
+  });
+  layerS1 = new S1GRDAWSEULayer({
+    dataProduct: '<data-product-id>',
+    acquisitionMode: AcquisitionMode.EW,
+    polarization: Polarization.DH,
+    resolution: Resolution.MEDIUM,
+  });
 ```
 
-It is also possible to create layers as they are defined in Sentinel Hub configuration instance:
+It is also possible to create layers by importing their definitions from the Sentinel Hub configuration instance:
 
 ```javascript
   import { LayersFactory } from '@sentinel-hub/sentinelhub-js';
@@ -103,23 +124,32 @@ The process of getting the authentication token is described in [Authentication 
 
 ## Fetching images
 
-Maps which correspond to these layers can be fetched via different protocols like WMS and Processing. Not all of the protocols can be used in all cases; for example, Processing can only render layers for which it has access to the `evalscript` and for which evalscript version 3 is used.
+Maps which correspond to these layers can be fetched via different protocols like WMS and Processing. Not all of the protocols can be used in all cases; for example, Processing can only render layers for which it has `evalscript` available and for which evalscript version 3 is used.
 
 ```javascript
   import { BBox, CRS_EPSG4326, MimeTypes, ApiType } from '@sentinel-hub/sentinelhub-js';
 
   const bbox = new BBox(CRS_EPSG4326, 18, 20, 20, 22);
-  const getMapParams = {
+  const fromTime = new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0));
+  const toTime = new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59));
+  const imageBlob = await layer.getMap({
     bbox: bbox,
-    fromTime: new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0)),
-    toTime: new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59)),
+    fromTime: fromTime,
+    toTime: toTime,
     width: 512,
     height: 512,
     format: MimeTypes.JPEG,
-  };
-
-  const imageBlob = await layer.getMap(getMapParams, ApiType.WMS);
-  const imageBlob2 = await layer.getMap(getMapParams, ApiType.PROCESSING);
+    api: ApiType.WMS,
+  });
+  const imageBlob2 = await layer.getMap({
+    bbox: bbox,
+    fromTime: fromTime,
+    toTime: toTime,
+    width: 512,
+    height: 512,
+    format: MimeTypes.JPEG,
+    api: ApiType.PROCESSING,
+  });
 ```
 
 Note that both of the images above should be _exactly_ the same.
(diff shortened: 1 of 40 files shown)


README.md
```diff
@@ -35,36 +35,57 @@ $ npm install @sentinel-hub/sentinelhub-js
...
+Such layer would only allow WMS requests. However, `Layer` is also a superclass for multiple dataset-specific subclasses (like `S1GRDAWSEULayer` - Sentinel-1 GRD data on AWS eu-central-1 Sentinel Hub endpoint) which can be instantiated with their own specific parameters and thus have additional capabilities.
 
 When it comes to Sentinel Hub layers, there are four ways to determine their content:
 
-- by `layerId`: ID of the layer, as used by OGC WMS and SentinelHub Configurator
+- by `instanceId` and `layerId`: ID of the layer, as used by OGC WMS and SentinelHub Configurator
 - by `evalscript`: custom (javascript) code that will be executed by the service per each pixel and will calculate the (usually RGB/RGBA) values
 - by `evalscriptUrl`: the URL from which the evalscript can be downloaded from
-- by `dataProduct`: the structure which contains an ID of a pre-existing product
+- by `dataProduct`: the ID of a pre-existing data product
```
Ardweaden: I think it would be appropriate to explain the difference between these approaches. Namely, that in the first case, only WMS can be used for `GetMap` etc. and only Processing for the rest. I am aware you mention this later, but it would be less confusing for a reader if explained here already.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -17,16 +27,16 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
+    instanceId = null,
+    layerId = null,
+    evalscript = null,
+    evalscriptUrl = null,
+    dataProduct = null,
+    title = null,
+    description = null,
+  }: ConstructorParameters) {
+    super({ title, description });
     if (layerId === null && evalscript === null && evalscriptUrl === null && dataProduct === null) {
```
Ardweaden: What happens if I specify a `layerId`, but not an `instanceId`? Unless I'm missing something, a check for that should be added as well 
` if (!(layerId !== null && linstanceId !== null ) || evalscript !== null || evalscriptUrl !== null && dataProduct !== null))`

README.md
grafolean: I see what you mean, but - the explanation is not that simple. InstanceId + layerId can be used for either WMS or Processing API (when authenticated) if the dataset supports it (because only SH v3 datasets support Processing API). If used with Processing API, configuration (`evalscript` and possibly other layer parameters) is fetched from service. `dataProduct` can only be used with SH v3, while evalscripts V3 are supported only on SH v3 (afaik)... and there are probably many more nuances. In other words, I don't have a good idea on how to improve this text without making it really complex... :) But maybe we can open a ticket there and try to word this a bit better? I agree that many users will have no idea what these 4 things are...

src/layer/AbstractSentinelHubV3Layer.ts
grafolean: Agree, good catch, we should fix that. It is however a bit of work, because we should check that for all of the layers - would you agree if we put it in a ticket and solve in a separate PR?

README.md
sinergise-anze: I see what you mean, but - the explanation is not that simple. InstanceId + layerId can be used for either WMS or Processing API (when authenticated) if the dataset supports it (because only SH v3 datasets support Processing API). If used with Processing API, configuration (`evalscript` and possibly other layer parameters) is fetched from service. `dataProduct` can only be used with SH v3, while evalscripts V3 are supported only on SH v3 (afaik)... and there are probably many more nuances. In other words, I don't have a good idea on how to improve this text without making it really complex... :) But maybe we can open a ticket there and try to word this a bit better? I agree that many users will have no idea what these 4 things are...

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: Agree, good catch, we should fix that. It is however a bit of work, because we should check that for all of the layers - would you agree if we put it in a ticket and solve in a separate PR?

Ardweaden: Of course, I see you already created it. Resolving.

README.md
Ardweaden: Make sense. Resolving.

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: Ah, nevermind... :) I started working on it and realized you were correct, the rest of the checks seem OK and this is the only change that needs to be done (I did write `if` a bit differently though). Unresolving this, fixed in this PR, please check... :)

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -7,22 +7,31 @@ import { GetMapParams, ApiType, PaginatedTiles } from 'src/layer/const';
 import { wmsGetMapUrl } from 'src/layer/wms';
 import { AbstractLayer } from 'src/layer/AbstractLayer';
 
+interface ConstructorParameters {
+  instanceId?: string | null;
+  layerId?: string | null;
+  evalscript?: string | null;
+  evalscriptUrl?: string | null;
+  title?: string | null;
+  description?: string | null;
```
zcernigoj: We should use "interface inheritance" ([docs](https://www.typescriptlang.org/docs/handbook/interfaces.html#extending-interfaces)) for the constructor parameters that are the same as in the super class / abstract class

sinergise-anze: Not sure about that... In theory, it is not necessarily true that the parameters of a child class include all the options of the parent class (though currently this is so, so maybe this is a moot point). Also, that would mean that we would need to export all the interfaces.

Is there any reason why we should do it?

stories/s1grdiw.eocloud.stories.js
```diff
@@ -41,17 +41,13 @@ export const getMapURL = () => {
...
-    null,
-    null,
-    null,
-    AcquisitionMode.IW,
-    Polarization.DV,
-    Resolution.HIGH,
-  );
+    acquisitionMode: AcquisitionMode.IW,
+    polarization: Polarization.DV,
+    resolution: Resolution.HIGH,
```
zcernigoj: `resolution` isn't a parameter for `S1GRDEOCloudLayer` anymore

stories/s1grdiw.eocloud.stories.js
```diff
@@ -77,17 +73,13 @@ export const getMapWMS = () => {
...
-      null,
-      null,
-      null,
-      AcquisitionMode.IW,
-      Polarization.DV,
-      Resolution.HIGH,
-    );
+      acquisitionMode: AcquisitionMode.IW,
+      polarization: Polarization.DV,
+      resolution: Resolution.HIGH,
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

stories/s1grdiw.eocloud.stories.js
```diff
@@ -146,19 +140,16 @@ export const getMapWMSEvalscript = () => {
...
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

stories/s1grdiw.eocloud.stories.js
```diff
@@ -177,17 +168,13 @@ export const getMapWMSEvalscript = () => {
...
-    null,
-    null,
-    null,
-    AcquisitionMode.IW,
-    Polarization.DV,
-    OrbitDirection.ASCENDING,
-  );
+    acquisitionMode: AcquisitionMode.IW,
+    polarization: Polarization.DV,
+    resolution: OrbitDirection.ASCENDING,
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

also: something went horribly wrong here :D

stories/s1grdiw.eocloud.stories.js
```diff
@@ -210,17 +197,13 @@ export const findTilesEPSG3857 = () => {
...
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

also same mistake as above (resolution: OrbitDirection.ASCENDING,)

stories/s1grdiw.eocloud.stories.js
```diff
@@ -283,17 +268,13 @@ export const findFlyovers = () => {
...
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

also same mistake as above (resolution: OrbitDirection.ASCENDING,)

stories/s1grdiw.eocloud.stories.js
```diff
@@ -335,17 +316,13 @@ export const findDatesEPSG4326 = () => {
...
```
zcernigoj: repeated just for easy find: resolution isn't a parameter for S1GRDEOCloudLayer anymore

also same mistake as above (resolution: OrbitDirection.ASCENDING,)

stories/s2l1c.stories.js
```diff
@@ -124,13 +124,14 @@ export const GetMapProcessing = () => {
 };
 
 export const GetMapWMSMaxCC20vs60 = () => {
-  const layerS2L1C20 = new S2L1CLayer(instanceId, s2l1cLayerId, null, null, null, null, null, 20);
-  const layerS2L1C60 = new S2L1CLayer(instanceId, s2l1cLayerId, null, null, null, null, null, 100);
+  const layerS2L1C20 = new S2L1CLayer({ instanceId, layerId, maxCloudCoverPercent: 20 });
+  const layerS2L1C60 = new S2L1CLayer({ instanceId, layerId, maxCloudCoverPercent: 100 });
```
zcernigoj: Sorry, that is my mistake from one of the previous PRs. `maxCloudCoverPercent` of the second one should be 60 or the name(s) should be fixed. Can you maybe fix it here?

stories/s3slstr.stories.js
```diff
@@ -192,7 +187,7 @@ export const findTiles = () => {
 };
 
 export const findFlyovers = () => {
-  const layer = new S3SLSTRLayer(instanceId, layerId, null, null, null, null, null, 60, 'NADIR');
+  const layer = new S3SLSTRLayer({ instanceId, layerId, maxCloudCoverPercent: 60, view: 'NADIR' });
```
zcernigoj: Would it be better to use `view.NADIR` (if available - similar to `orbitDirection` and others in S1 GRD layers) instead of using string?

src/layer/AbstractSentinelHubV1OrV2Layer.ts
zcernigoj: > Is there any reason why we should do it?

I just have one: They are repeated ... a number of times and I am too lazy to check if some param from the `Abstract...Layer` classes is missing in the other `...Layer` classes :)

> Edit: I was actually expecting a question why I used `interface`s instead of `type`s... ;)

It looks like a nice solution so I don't have a reason to question the use of `interface`s

stories/s1grdiw.eocloud.stories.js
sinergise-anze: Thanks, there was some mix-up (`resolution: OrbitDirection.ASCENDING`) with parameters too, so I changed all of them to `orbitDirection`. Fixed.

stories/s2l1c.stories.js
sinergise-anze: Sure, though it still doesn't work (wrong dates - 2014). But that should probably be solved in another PR.

zcernigoj: Thanks! I know, but just to put the `maxCloudCoverPercent` out of the mind. I agree about the dates.

stories/s3slstr.stories.js
sinergise-anze: Good idea, thx, fixed.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
sinergise-anze: I see your point, and I do agree to some extent. The problem is just that it seems conceptually wrong to me... For example, AbstractSH3Layer could just as easily be a subclass of WmsLayer (it is not, but for the sake of the argument - and there are reasons why it should be). In this case you would have `baseUrl` in the parent class, but only `instanceId` in the child class (`baseUrl` is constructed from dataset-specific intrinsic knowledge + `instanceId`).

I think it makes sense not to extend constructor parameters tbh... In any case, can we maybe leave it now as it is and then try to find agreement live?

stories/s1grdiw.eocloud.stories.js
zcernigoj: Nice, resolving.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
zcernigoj: No problem, it can stay like that. Resolving.

stories/s3slstr.stories.js
zcernigoj: Nice, but here on the line 189 it's still a string.

stories/s2l1c.stories.js
```diff
@@ -54,7 +54,7 @@ export const GetMapWMS = () => {
 
   // getMap is async:
   const perform = async () => {
-    const layerS2L1C = new S2L1CLayer(instanceId, s2l1cLayerId);
+    const layerS2L1C = new S2L1CLayer({ instanceId, layerId: layerId });
```
zcernigoj: `layerId: layerId` could be changed to just `layerId`

stories/s3slstr.stories.js
sinergise-anze: Thx, overlooked, fixed.

stories/s2l1c.stories.js
sinergise-anze: Fixed.

stories/s3slstr.stories.js
zcernigoj: Great, resolving.

stories/s2l1c.stories.js
zcernigoj: Great, resolving.

PR #48, opened 2020-03-16
Support findDates() on WmsLayer

sinergise-anze:
This PR implements `findDates()` for WMS (tested on Proba-V). Since we only need to support a small subset of [specification](http://cite.opengeospatial.org/OGCTestData/wms/1.1.1/spec/wms1.1.1.html#dims). Note that WmsLayer's `findDates()` throws an exception whenever it encounters (as yet) unsupported format.

### DIFF
--- a/src/layer/WmsLayer.ts
+++ b/src/layer/WmsLayer.ts
@@ -1,6 +1,10 @@
+import moment from 'moment';
+
+import { BBox } from 'src/bbox';
 import { GetMapParams, ApiType } from 'src/layer/const';
 import { wmsGetMapUrl } from 'src/layer/wms';
 import { AbstractLayer } from 'src/layer/AbstractLayer';
+import { LayersFactory } from 'src/layer/LayersFactory';
 
 interface ConstructorParameters {
   baseUrl?: string;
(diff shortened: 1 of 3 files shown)


src/layer/WmsLayer.ts
```diff
@@ -26,4 +30,40 @@ export class WmsLayer extends AbstractLayer {
...
+      throw new Error('Layer time information is not in ISO8601 format, parsing not supported');
+    }
+    const times = timeDimension['_'];
+    if (!times.includes(',')) {
+      throw new Error(
+        'Currently only time lists of length > 1 are supported (valid time values separated by comma)',
+      );
+    }
+
+    const result = times.split(',').map((t: string) => moment.utc(t).toDate());
```
zcernigoj: Did you mean to say that periods are not supported?

Because in [GIBS WMS](https://gibs.earthdata.nasa.gov/wms/epsg3857/best/wms.cgi?SERVICE=WMS&request=GetCapabilities), there is the example below, that would not work.

```
<Dimension name="time" units="ISO8601" default="2020-03-02" nearestValue="1">
2012-07-05/2012-12-27/P5D,2013-01-01/2013-12-27/P5D,2014-01-01/2014-12-27/P5D,2015-01-01/2015-12-27/P5D,2016-01-01/2016-02-20/P5D,2016-02-25/2016-03-02/P6D,2016-03-02/2016-12-27/P5D,2017-01-01/2017-12-27/P5D,2018-01-01/2018-12-27/P5D,2019-01-01/2019-12-27/P5D,2020-01-01/2020-02-20/P5D,2020-02-25/2020-03-02/P6D,2020-03-02/2020-03-02/P5D
</Dimension>
```

sinergise-anze: Ah, interesting... Good point, let me fix that. :)

zcernigoj: To correct myself, maybe it would work

> As of 2.3.0, moment also supports parsing ISO 8601 durations.

[moment docs for duration](https://momentjs.com/docs/#/durations/creating/)

sinergise-anze: As discussed live, it turns out that there are some other modifications needed to make GIBS work, so I'll fix this in another PR.

zcernigoj: Ok, resolving.

PR #49, opened 2020-03-17
Fix/gibs

sinergise-anze:
A few changes were needed to support GIBS:

- `fetchGetCapabilities*()` are now in `utils.js` (to avoid circular imports - `LayersFactory -> WmsLayer -> LayersFactory`)
- fixed: `fetchGetCapabilitiesJson` is internal to the library (was even private to LayersFactory) and as such doesn't belong to node.js example; we want users to use our abstraction layers, if they want raw access, they should use fetch / axios / ... directly
- fixed: GIBS doesn't allow `from/to` format of dates; allow setting `fromDate` to `null`, in which case only one date is passed on
- fixed: GIBS doesn't allow milliseconds in dates, changed date formatting (other known services also support `YYYY-MM-DDThh:mm:ssZ` format which is now used)
- fixed: since we are using WMS version `1.1.1`, we should be using `srs` instead of `crs` parameter (other services don't mind, but GIBS follows the standard here)
- added support for `start/end/duration` times in GetCapabilities layer time dimension

### DIFF
--- a/stories/wms.gibs.stories.js
+++ b/stories/wms.gibs.stories.js
@@ -183,14 +183,13 @@ export const findDates = () => {
     const dates = await layer.findDates(bbox, fromTime, toTime);
     containerEl.innerHTML = JSON.stringify(dates, null, true);
 
-    const resDateStartOfDay = new Date(new Date(dates[0]).setUTCHours(0, 0, 0, 0));
-    const resDateEndOfDay = new Date(new Date(dates[0]).setUTCHours(23, 59, 59, 999));
+    const resDateStartOfDay = new Date(new Date(dates[5]).setUTCHours(0, 0, 0, 0));
 
     // prepare an image to show that the number makes sense:
     const getMapParams = {
       bbox: bbox4326,
-      fromTime: resDateStartOfDay,
-      toTime: resDateEndOfDay,
+      fromTime: null,
+      toTime: resDateStartOfDay,
       width: 512,
       height: 512,
       format: MimeTypes.JPEG,
--- a/stories/wms.gibs.stories.js
+++ b/stories/wms.gibs.stories.js
@@ -11,7 +11,6 @@ import {
 } from '../dist/sentinelHub.esm';
 
 const baseUrl = 'https://gibs.earthdata.nasa.gov/wms/epsg4326/best/wms.cgi';
-// const layerId = 'AIRS_L2_Carbon_Monoxide_500hPa_Volume_Mixing_Ratio_Night';
 const layerId = 'MODIS_Terra_SurfaceReflectance_Bands721';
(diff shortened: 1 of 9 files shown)


stories/wms.gibs.stories.js
```diff
@@ -0,0 +1,203 @@
...
+  CRS_EPSG3857,
+  CRS_EPSG4326,
+  BBox,
+  MimeTypes,
+  ApiType,
+  LayersFactory,
+} from '../dist/sentinelHub.esm';
+
+const baseUrl = 'https://gibs.earthdata.nasa.gov/wms/epsg4326/best/wms.cgi';
+// const layerId = 'AIRS_L2_Carbon_Monoxide_500hPa_Volume_Mixing_Ratio_Night';
```
zcernigoj: Should this be removed?

sinergise-anze: I left it there as a different layer, if someone wants to test... But yeah, it probably doesn't make sense to leave it there, removed, thx.

zcernigoj:
> fixed: GIBS doesn't allow from/to format of dates; allow setting fromDate to null, in which case only one date is passed on

I think that allowing to only set the `toTime` if the datasource doesn't accept `fromTime/toTime` should be described in the readme as well.

sinergise-anze:
> > fixed: GIBS doesn't allow from/to format of dates; allow setting fromDate to null, in which case only one date is passed on
> 
> I think that allowing to only set the `toTime` if the datasource doesn't accept `fromTime/toTime` should be described in the readme as well.

You have a point, this behavior is a bit surprising. However I am not sure it deserves a place in readme (which serves more as a User Guide), so I have documented it in TypeDoc instead and would appreciate your opinion.

I am aware that this documentation is currently buried very deep and users won't be able to find it (they need to run `npm run doc-build` and navigate browser to `doc/index.html`, then find the relevant part there), but at least this way we won't for get to add it when we improve API documentation.

Do you think this is a good approach?

stories/wms.gibs.stories.js
zcernigoj: We can create a new story for some different layer someday :)
Resolving.

zcernigoj:
> > > fixed: GIBS doesn't allow from/to format of dates; allow setting fromDate to null, in which case only one date is passed on
> > 
> > 
> > I think that allowing to only set the `toTime` if the datasource doesn't accept `fromTime/toTime` should be described in the readme as well.
> 
> You have a point, this behavior is a bit surprising. However I am not sure it deserves a place in readme (which serves more as a User Guide), so I have documented it in TypeDoc instead and would appreciate your opinion.
> 
> I am aware that this documentation is currently buried very deep and users won't be able to find it (they need to run `npm run doc-build` and navigate browser to `doc/index.html`, then find the relevant part there), but at least this way we won't for get to add it when we improve API documentation.
> 
> Do you think this is a good approach?

As discussed live, it's not optimal, but it's a good approach. Resolving.

PR #50, opened 2020-03-18
Fix node.js examples

sinergise-anze:
- use new constructor function signature in node.js examples
- if examples fail, exit status should be non-zero

### DIFF
--- a/example/node/index.js
+++ b/example/node/index.js
@@ -82,7 +82,7 @@ async function run() {
 
   // layer can also be constructed directly:
   const layerId = layers[0].layerId; // take a layer that we know exists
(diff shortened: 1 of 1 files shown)


example/node/index.js
```diff
@@ -88,7 +88,7 @@ async function run() {
```
dgostencnik: WmsLayer should be also changed (couldn't put comment on exact line)
 
const layer = new WmsLayer({ baseUrl, layerId, title: 'My title', description: 'My desc' });

sinergise-anze: Nice catch, thank you! Fixed.

I have checked now if linter would help protect against such errors, but couldn't make it issue a warning when the number of parameters was wrong, so I didn't enable it. I did however remove unused `OrbitDirection` in the process.

PR #51, opened 2020-03-18
Rename findDates() to findDatesUTC()

sinergise-anze:
To avoid surprising users, we should rename `findDates` to `findDatesUTC`. 

This PR also provides a (deprecated) `findDates` method which issues a warning and calls `findDatesUTC`. I added a story to test `findDates`, but removed it later to avoid "polluting" the storybook.

### DIFF
--- a/src/layer/AbstractSentinelHubV1OrV2Layer.ts
+++ b/src/layer/AbstractSentinelHubV1OrV2Layer.ts
@@ -119,8 +119,8 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
     return {};
   }
 
-  public async findDates(bbox: BBox, fromTime: Date, toTime: Date): Promise<Date[]> {
-    if (!this.dataset.findDatesUrl) {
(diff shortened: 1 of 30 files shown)


src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -131,7 +131,7 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
       ...this.getFindDatesAdditionalParameters(),
```
zcernigoj: Should `getFindDatesAdditionalParameters()` also be renamed (to `getFindDatesUTCAdditionalParameters` I suppose)?

sinergise-anze: True, thank you, fixed.

zcernigoj: Nice, resolving.

PR #52, opened 2020-03-18
Do not send auth token to external services

sinergise-anze:
Due to a bug (old code) in `fetchCached()`, which indiscriminately added auth token to any request, library was sending auth token to external services. This PR solves this by replacing `fetchCached` with `axios.get()` + cache interceptors.

Because the tests were complaining about invalid fields in `AxiosRequestConfig`, we [monkey patch](https://stackoverflow.com/a/58778671/593487) the type (officially known as [Declaration Merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).

Consequently, `RequestConfig` was no longer needed as we can now use `AxiosRequestConfig` directly.

Since `forceFetch` parameter was not used anywhere, it was also removed.

### DIFF
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosRequestConfig } from 'axios';
+import axios from 'axios';
 import { stringify } from 'query-string';
 
 const SENTINEL_HUB_CACHE = 'sentinelhub-v1';
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -30,8 +30,8 @@ const fetchCachedResponse = async (request: any): Promise<any> => {
   if (typeof window === 'undefined' || !window.caches) {
     return request;
   }
-  // resource not cacheable? It couldn't have been saved to cache:
   const cacheKey = await generateCacheKey(request);
+  // resource not cacheable? It couldn't have been saved to cache:
   if (cacheKey === null) {
     return request;
   }
@@ -89,12 +89,11 @@ const saveCacheResponse = async (response: any): Promise<any> => {
   if (typeof window === 'undefined' || !window.caches) {
     return response;
   }
-  // resource not cacheable?
   const cacheKey = await generateCacheKey(response.config);
+  // resource not cacheable?
   if (cacheKey === null) {
     return response;
   }
-  // Cache API not supported?
   let cache;
   try {
     cache = await caches.open(SENTINEL_HUB_CACHE);
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -68,7 +68,6 @@ const fetchCachedResponse = async (request: any): Promise<any> => {
         responseData = await cachedResponse.text();
         break;
       case 'json':
-      case undefined:
         responseData = await cachedResponse.json();
         break;
       default:
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -19,7 +19,7 @@ declare module 'axios' {
 export const registerAxiosCacheRetryInterceptors = (): any => {
   findAndDeleteExpiredCachedItems();
   axios.interceptors.request.use(fetchCachedResponse, error => Promise.reject(error));
-  axios.interceptors.response.use(setCacheResponse, retryRequests);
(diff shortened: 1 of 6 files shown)


zcernigoj:
Also: `Uncaught (in promise) SyntaxError: Unexpected token o in JSON at position 1` when using authentication

src/utils/axiosInterceptors.ts
```diff
@@ -6,9 +6,14 @@ const EXPIRY_HEADER_KEY = 'Cache_Expires';
...
 const DELAY = 3000;
 const RETRIES = 5;
-export interface RequestConfig extends AxiosRequestConfig {
-  useCache?: boolean;
-  retries?: number;
+
+// we are extending axios' AxiosRequestConfig:
+// https://stackoverflow.com/questions/58777924/axios-typescript-customize-axiosrequestconfig
+declare module 'axios' {
+  export interface AxiosRequestConfig {
```
zcernigoj: VS Code says the `AxiosRequestConfig` in `import axios, { AxiosRequestConfig } from 'axios'; is not used.
Lint doesn't say anything. I tried without importing `AxiosRequestConfig` it also works. 

Don't know what to do here. Should it be removed from the import?

sinergise-anze: Nice catch! Yes, it should be removed, because we are not really using it - interesting that linter didn't complain. Thanks, fixed.

zcernigoj: Nice. Resolving.

sinergise-anze:
> Also: `Uncaught (in promise) SyntaxError: Unexpected token o in JSON at position 1` when using authentication in BYOC, Landsat 8, ... stories

Thank you for catching this! It turns out that caching of JSON resources didn't work properly, because axios performed transformation from JSON string to object twice. I couldn't find a better way to solve this, so we now convert from object to JSON string right before saving to cache. Not the cleanest way, but... it works.

sinergise-anze:
> When running the node example, this pops up multiple times: `Caching failed ReferenceError: caches is not defined`
> 
> `caches` occurs in `src/utils/axiosInterceptors.ts`

Yes - Cache API is not supported in node.js. Not sure which solution should be used instead, so now caching is simply (silently) skipped if not supported. Not the best solution, but we can deal with this if/when need arises.

src/utils/axiosInterceptors.ts
```diff
@@ -6,36 +6,34 @@ const EXPIRY_HEADER_KEY = 'Cache_Expires';
...
-    case 'get':
-      cacheKey = generateCacheKey(request.url, stringify(request.params));
-      break;
-    default:
-      return request;
+  // do not perform caching if Cache API is not supported:
+  if (typeof window === 'undefined' || !window.caches) {
+    return request;
+  }
+  // resource not cacheable? It couldn't have been saved to cache:
```
zcernigoj: Maybe move this comment after `const cacheKey = await generateCacheKey(request);` so it wouldn't look like `cacheKey ` is used just for checking if the item is *not* cached.

src/utils/axiosInterceptors.ts
```diff
@@ -82,38 +80,54 @@ const fetchCachedResponse = async (request: any): Promise<any> => {
...
+  // do not perform caching if Cache API is not supported:
+  if (typeof window === 'undefined' || !window.caches) {
+    return response;
+  }
+  // resource not cacheable?
+  const cacheKey = await generateCacheKey(response.config);
+  if (cacheKey === null) {
+    return response;
+  }
+  // Cache API not supported?
```
zcernigoj: I think this doesn't belong here; on line 88, there's also a similar comment `  // do not perform caching if Cache API is not supported:`

src/utils/axiosInterceptors.ts
sinergise-anze: Thanks, good point! Fixed.

src/utils/axiosInterceptors.ts
sinergise-anze: :+1:, fixed.

src/utils/axiosInterceptors.ts
zcernigoj: Great. Resolving.

src/utils/axiosInterceptors.ts
zcernigoj: Resolving.

PR #53, opened 2020-03-18
fix refactored LayerFactory problems

zcernigoj:
After a small refactor of `src/layer/LayerFactory.ts` in https://github.com/sentinel-hub/sentinelhub-js/pull/49, `parseSHInstanceId()` was still in `src/layer/LayerFactory.ts`, but was called as `this.parseSHInstanceId()` in `src/layer/utils.ts`. So I moved it to `src/layer/utils.ts`

I also moved `SH_SERVICE_HOSTNAMES_V1_OR_V2`, `SH_SERVICE_HOSTNAMES_V3` from `src/layer/LayersFactory.ts` to `src/layer/const.ts` as it seems to be the right place for them, and they are needed both in  `src/layer/LayerFactory.ts` and `src/layer/utils.ts`.

Other changes are from running `npm run build` (which somehow reformats `package.json`) and `npm run prettier`.

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -4,7 +4,7 @@ import {
   fetchGetCapabilitiesJson,
   parseSHInstanceId,
 } from 'src/layer/utils';
-import { SH_SERVICE_HOSTNAMES_V1_OR_V2, SH_SERVICE_HOSTNAMES_V3 } from './const';
+import { SH_SERVICE_HOSTNAMES_V1_OR_V2, SH_SERVICE_HOSTNAMES_V3 } from 'src/layer/const';
 import {
   DATASET_S2L2A,
   DATASET_AWS_L8L1C,
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -2,7 +2,9 @@ import {
   fetchGetCapabilitiesXml,
   fetchGetCapabilitiesJsonV1,
   fetchGetCapabilitiesJson,
+  parseSHInstanceId,
 } from 'src/layer/utils';
+import { SH_SERVICE_HOSTNAMES_V1_OR_V2, SH_SERVICE_HOSTNAMES_V3 } from './const';
 import {
   DATASET_S2L2A,
   DATASET_AWS_L8L1C,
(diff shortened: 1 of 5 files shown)


src/layer/LayersFactory.ts
```diff
@@ -46,14 +48,6 @@ export class LayersFactory {
...
 
-  public static readonly SH_SERVICE_HOSTNAMES_V1_OR_V2: string[] = ['https://eocloud.sentinel-hub.com/'];
-
-  public static readonly SH_SERVICE_HOSTNAMES_V3: string[] = [
-    'https://services.sentinel-hub.com/',
-    'https://services-uswest2.sentinel-hub.com/',
-    'https://creodias.sentinel-hub.com/',
-  ];
-
   private static readonly DATASET_FROM_JSON_GETCAPAPABILITIES = {
```
zcernigoj: 1. This should be named `DATASET_FROM_JSON_GETCAPAPABILITIES_V3`

2. Should this and `DATASET_FROM_JSON_GETCAPABILITIES_V1`, `LAYER_FROM_DATASET_V3`, `LAYER_FROM_DATASET_V12` also be moved to `./const.ts`?

src/layer/utils.ts
```diff
@@ -3,6 +3,7 @@ import { stringify } from 'query-string';
 import { parseStringPromise } from 'xml2js';
 
 import { getAuthToken, isAuthTokenSet } from 'src/auth';
+import { SH_SERVICE_HOSTNAMES_V1_OR_V2, SH_SERVICE_HOSTNAMES_V3 } from './const';
```
zcernigoj: Change `./const` to `src/const`

src/layer/LayersFactory.ts
```diff
@@ -2,7 +2,9 @@ import {
   fetchGetCapabilitiesXml,
   fetchGetCapabilitiesJsonV1,
   fetchGetCapabilitiesJson,
+  parseSHInstanceId,
 } from 'src/layer/utils';
+import { SH_SERVICE_HOSTNAMES_V1_OR_V2, SH_SERVICE_HOSTNAMES_V3 } from './const';
```
zcernigoj: Change `./const` to `src/const`

src/layer/LayersFactory.ts
sinergise-anze: 1. Agree, not critical though

2. Not sure, tbh... Maybe we can leave it here for now? If it bothers we can always move it later.

src/layer/utils.ts
zcernigoj: Fixed

src/layer/LayersFactory.ts
zcernigoj: Fixed.

src/layer/LayersFactory.ts
zcernigoj: Ok, I will leave that for later. Resolving

PR #54, opened 2020-03-19
fix / WMSLayer.findDatesUTC() should use params

zcernigoj:
WMSLayer.findDatesUTC() should return only dates for requested time interval.

### DIFF
--- a/src/layer/WmsLayer.ts
+++ b/src/layer/WmsLayer.ts
@@ -33,8 +33,8 @@ export class WmsLayer extends AbstractLayer {
 
   public async findDatesUTC(
     bbox: BBox, // eslint-disable-line @typescript-eslint/no-unused-vars
-    fromTime: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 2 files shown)


src/layer/WmsLayer.ts
```diff
@@ -56,26 +56,30 @@ export class WmsLayer extends AbstractLayer {
...
           break;
         case 3:
-          const [fromTime, toTime, interval] = timeParts;
-          const intervalDuration = moment.duration(interval);
-          const toTimeMoment = moment.utc(toTime);
-          for (let t = moment.utc(fromTime); t.isSameOrBefore(toTimeMoment); t.add(intervalDuration)) {
-            result.push(t.toDate());
+          const [timePartFromTime, timePartToTime, timePartInterval] = timeParts;
+          const fromTimeMoment = moment.utc(timePartFromTime);
+          const toTimeMoment = moment.utc(timePartToTime);
```
zcernigoj: Should these two be renamed - do they seem too associated with input params `fromTime` and `toTime`?

zcernigoj: Renamed

PR #57, opened 2020-03-23
Node.js doesn't support btoa()

sinergise-anze:
Since node.js doesn't support `btoa`, we need to use a workaround. The example that I added to `examples/node/index.js` shows that it now works.

### DIFF
--- a/src/legacyCompat.ts
+++ b/src/legacyCompat.ts
@@ -65,7 +65,7 @@ export async function legacyGetMapFromParams(
             decodedEvalscript = atob(evalscript);
           } else {
             // node.js doesn't support atob:
-            decodedEvalscript = Buffer.from(evalscript, 'base64').toString('ascii');
+            decodedEvalscript = Buffer.from(evalscript, 'base64').toString('utf8');
           }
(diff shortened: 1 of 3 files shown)


src/legacyCompat.ts
```diff
@@ -60,7 +60,13 @@ export async function legacyGetMapFromParams(
           throw new Error('Processing API is only possible with SH V3 layers');
         }
         if (evalscript) {
-          const decodedEvalscript = atob(evalscript);
+          let decodedEvalscript;
+          if (typeof window !== 'undefined' && window.atob) {
+            decodedEvalscript = atob(evalscript);
+          } else {
+            // node.js doesn't support atob:
+            decodedEvalscript = Buffer.from(evalscript, 'base64').toString('ascii');
```
zcernigoj: Should it be `utf-8` instead of `ascii`?

sinergise-anze: Agreed, probably best to use UTF-8 and be explicit about it (I have now added encoding to "`btoa`" too).

zcernigoj: Ok, resolving.

PR #60, opened 2020-04-01
Feature/data fusion

sinergise-anze:
This PR should be merged as soon as possible so that it can be used in EO Browser. To that end, there are a few things that I have avoided doing and which should be fixed at a later date:

- layer parameters (such as maxCloudCoverPercent) are not supported yet; to do that, each layer should be concerned with creating its own part of Processing request (future PR)
- mosaickingOrder is (still) hardcoded (missing both from GetMapParams and DataFusionLayerInfo)
- data fusion would probably deserve a mention (and example) in README.md, but I have trouble finding public documentation and examples at this point (understandable, since functionality is still very new)

### DIFF
--- a/stories/datafusion.stories.js
+++ b/stories/datafusion.stories.js
@@ -1,0 +1,110 @@
+import { renderTilesList, setAuthTokenWithOAuthCredentials } from './storiesUtils';
+
+import {
+  ProcessingDataFusionLayer,
+  CRS_EPSG3857,
+  BBox,
+  MimeTypes,
+  ApiType,
+  S2L1CLayer,
+  S2L2ALayer,
+} from '../dist/sentinelHub.esm';
+
+if (!process.env.INSTANCE_ID) {
+  throw new Error('INSTANCE_ID environment variable is not defined!');
+}
+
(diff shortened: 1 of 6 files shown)


stories/datafusion.stories.js
```diff
@@ -0,0 +1,110 @@
...
+if (!process.env.S2L1C_LAYER_ID) {
+  throw new Error('S2L1C_LAYER_ID environment variable is not defined!');
+}
+if (!process.env.S2L2A_LAYER_ID) {
+  throw new Error('S2L2A_LAYER_ID environment variable is not defined!');
+}
+
+const instanceId = process.env.INSTANCE_ID;
+const layerS2L1CId = process.env.S2L1C_LAYER_ID;
+const layerS2L2AId = process.env.S2L2A_LAYER_ID;
```
zcernigoj, line 26: If it helps, in `index.stories.js` these names are formulated in a different order (`s2l2aLayerId`)

sinergise-anze: Ok, renamed.

zcernigoj: Ok, resolving.

PR #61, opened 2020-04-03
Feature/debug

sinergise-anze:
This PR implements construction of `curl` commands and their logging in console, which allows for easier debugging of failed requests. 

Usage:
```javascript
setDebugEnabled(true);
const x = await layer.getMap(params, ApiType.WMS);
setDebugEnabled(false);
```

Note: this is a different functionality from what we discussed (`getMapCurl`), though `curlify` function could be used there too.


src/layer/utils.ts
```diff
@@ -80,3 +80,9 @@ export function parseSHInstanceId(baseUrl: string): string {
   }
   throw new Error(`Could not parse instanceId from URL: ${baseUrl}`);
 }
+
+export let debugEnabled = false;
```
dgostencnik, line 84: This could probably be moved to some better location as this debug switch is not limited to layers only, but it can be used for all axios request. fftr

sinergise-anze: Good point! I have moved it now to `utils/debug.ts`. 

While at it, I have also removed access to `debugEnabled` variable and implemented a (trivial) getter.

PR #62, opened 2020-04-03
Feature/stats and histogram

guss84:
Add `getStatsAndHistogram` method

Things to note.
 - Eocloud only supports get requests
 - v3 evalscripts are not supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
   - [Searching for data](#searching-for-data)
   - [Backwards compatibility](#backwards-compatibility)
   - [Authentication for Processing API](#authentication-for-processing-api)
+  - [Debugging](#debugging)
 - [Examples](#examples)
   - [Preparation before running examples](#preparation-before-running-examples)
   - [Running examples](#running-examples)
@@ -238,6 +239,20 @@ const clientSecret = /* OAuth client's secret, best to put it in .env file and u
 const authToken = await requestAuthToken(clientId, clientSecret);
 setAuthToken(authToken);
 ```
+
+## Debugging
+
+This library is an abstraction layer that provides nice interface for accessing the underlying services, which simplifies development - but when requests fail, it is sometimes difficult to understand why. To enable easier debugging, `setDebugEnabled` can be used:
+
+```javascript
+import { setDebugEnabled } from '@sentinel-hub/sentinelhub-js';
+
+setDebugEnabled(true);
+// ... failing operation
+setDebugEnabled(false);
+```
+
+While debug mode is enabled, library will output any request it makes (even if the response comes from cache) to console in the form of a `curl` command.
 
 # Examples
 This project contains some examples to demonstrate how the library is used.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
   - [Fetching images](#fetching-images)
   - [Searching for data](#searching-for-data)
   - [Backwards compatibility](#backwards-compatibility)
-- [Authentication for Processing API](#authentication-for-processing-api)
+  - [Authentication for Processing API](#authentication-for-processing-api)
 - [Examples](#examples)
   - [Preparation before running examples](#preparation-before-running-examples)
   - [Running examples](#running-examples)
@@ -219,7 +219,7 @@ If we already have a WMS GetMap URL, we can use it directly:
   const imageBlob4 = await legacyGetMapFromUrl(fullUrlWithWmsQueryString, ApiType.PROCESSING);
 ```
 
-# Authentication for Processing API
+## Authentication for Processing API
 
 Requests to Processing API need to be authenticated.
 Documentation about authentication is available at [Sentinel Hub documentation](https://docs.sentinel-hub.com/api/latest/#/API/authentication).
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ const stats = await layerS2L2A.getStats({
   fromTime: fromTime, 
   toTime: toTime, 
   resolution: resolution,
-  binsAmount:10,
+  bins:10,
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ We can always use layer to search for data availability:
 
 ## Getting basic statistics and histogram
 
-Getting basic statistics(mean, min, max, standard deviation) and a histogram for a geometry(Polygon or MultiPolygon).
+Getting basic statistics (mean, min, max, standard deviation) and a histogram for a geometry (Polygon or MultiPolygon).
 The histogram uses the `equalfrequency` binning method and defaults to 5 bins.
 
 ```javascript
--- a/README.md
+++ b/README.md
@@ -180,6 +180,29 @@ We can always use layer to search for data availability:
   const datesS1 = await layerS1.findDatesUTC(bbox, fromTime, toTime);
 ```
 
+## Getting basic statistics and histogram
+
+Getting basic statistics(mean, min, max, standard deviation) and a histogram for a geometry(Polygon or MultiPolygon) on a specific date.
+The histogram uses the `equalfrequency` binning method and defaults to 5 bins.
+
+```javascript
+  const layerS2L2A = new S2L2ALayer({
+    instanceId: '<my-instance-id>',
+    layerId: 'LAYER_S2L2A',
+    maxCloudCoverPercent: 100,
+  });
+
+const stats = await layerS2L2A.getStatsAndHistogram({
+  geometry: geometry, 
+  crs: CRS_EPSG4326, 
+  fromTime: fromTime, 
+  toTime: toTime, 
+  resolution: resolution,
+  binsAmount:10,
+});
+```
+
+
 ## Backwards compatibility
 
 To make it easier to use this library with legacy code, there are two functions that are implemented on top of the library, which do not require instantiating a `Layer` subclass.
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ The histogram uses the `equalfrequency` binning method and defaults to 5 bins.
     maxCloudCoverPercent: 100,
   });
 
-const stats = await layerS2L2A.getStatsAndHistogram({
+const stats = await layerS2L2A.getStats({
   geometry: geometry, 
   crs: CRS_EPSG4326, 
   fromTime: fromTime, 
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ const stats = await layerS2L2A.getStats({
   fromTime: fromTime, 
   toTime: toTime, 
   resolution: resolution,
-  bins:10,
+  bins: 10,
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -194,7 +194,6 @@ The histogram uses the `equalfrequency` binning method and defaults to 5 bins.
 
 const stats = await layerS2L2A.getStats({
   geometry: geometry, 
-  crs: CRS_EPSG4326, 
   fromTime: fromTime, 
   toTime: toTime, 
   resolution: resolution,
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ We can always use layer to search for data availability:
 
 ## Getting basic statistics and histogram
 
-Getting basic statistics(mean, min, max, standard deviation) and a histogram for a geometry(Polygon or MultiPolygon) on a specific date.
+Getting basic statistics(mean, min, max, standard deviation) and a histogram for a geometry(Polygon or MultiPolygon).
 The histogram uses the `equalfrequency` binning method and defaults to 5 bins.
 
 ```javascript
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -242,5 +242,10 @@ export class AbstractLayer {
     return await this.findDatesUTC(bbox, fromTime, toTime);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async getStatsAndHistogram(payload: any): Promise<any> {
+    throw new Error('getStatsAndHistogram() not implemented for this dataset');
+  }
+
   public async updateLayerFromServiceIfNeeded(): Promise<void> {}
 }
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -1,13 +1,12 @@
 import axios, { AxiosRequestConfig } from 'axios';
 import moment from 'moment';
-
-import { GetMapParams, ApiType, PaginatedTiles, FlyoverInterval } from 'src/layer/const';
-import { BBox } from 'src/bbox';
-import { Dataset } from 'src/layer/dataset';
 import area from '@turf/area';
 import { union, intersection, Geom } from 'polygon-clipping';
 
+import { BBox } from 'src/bbox';
 import { CRS_EPSG4326 } from 'src/crs';
+import { GetMapParams, ApiType, PaginatedTiles, FlyoverInterval } from 'src/layer/const';
+import { Dataset } from 'src/layer/dataset';
 
 interface ConstructorParameters {
   title?: string | null;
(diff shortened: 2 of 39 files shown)


stories/landsat5.eocloud.stories.js
```diff
@@ -251,3 +251,33 @@ export const findDatesUTCEPSG3857 = () => {
...
+  wrapperEl.innerHTML = `<h2>getStatsAndHistogram</h2>`;
+
+  const containerEl = document.createElement('pre');
+  wrapperEl.insertAdjacentElement('beforeend', containerEl);
+
+  const fromTime = new Date(Date.UTC(2000, 1 - 1, 1, 0, 0, 0));
+  const toTime = new Date(Date.UTC(2000, 3 - 1, 15, 23, 59, 59));
+
+  const perform = async () => {
+    const payload = {
```
sinergise-anze, line 269: Nitpick: `payload` is usually something that we send somewhere, maybe we should use `params` here to avoid confusing readers?

EDIT: in other stories too...

src/layer/const.ts
```diff
@@ -99,3 +100,54 @@ export const SH_SERVICE_HOSTNAMES_V3: string[] = [
...
+  type?: HistogramType;
+  bins?: number;
+  evalscript?: string;
+  evalsource?: string;
+};
+
+export enum HistogramType {
+  EQUALFREQUENCY = 'EQUALFREQUENCY',
+  EQUIDISTANT = 'equidistant',
+  STREAMING = 'streaming',
```
sinergise-anze, line 129: I would suggest removing these two options, since they are not supported (alternatively, we can also just comment them out - but maybe then they should be uppercased for consistency?), to avoid confusion.

README.md
```diff
@@ -180,6 +180,29 @@ We can always use layer to search for data availability:
...
+    maxCloudCoverPercent: 100,
+  });
+
+const stats = await layerS2L2A.getStatsAndHistogram({
+  geometry: geometry, 
+  crs: CRS_EPSG4326, 
+  fromTime: fromTime, 
+  toTime: toTime, 
+  resolution: resolution,
+  binsAmount:10,
```
sinergise-anze, line 201: Nitpick, fftr: `binsAmount` sounds weird to me, maybe `nBins` or just `bins`?

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -140,4 +155,53 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+      throw new Error('Parameter "geometry" needs to be provided');
+    }
+    if (!params.crs) {
+      throw new Error('Parameter "crs" needs to be provided');
+    }
+    if (!params.resolution) {
+      throw new Error('Parameter "resolution" needs to be provided');
+    }
+    if (!params.fromTime || !params.toTime) {
+      throw new Error('Parameter "fromTime" and "toTime" needs to be provided');
```
sinergise-anze, line 170: `'Parameters "fromTime" and "toTime" need to be provided'`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -140,4 +155,53 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+    }
+
+    const { data } = await axios.get(this.dataset.shServiceHostname + 'v1/fis/' + this.instanceId, {
+      params: payload,
+    });
+    // convert date strings to Date objects
+    Object.keys(data).forEach(key => {
+      data[key].forEach((dailyStats: StatsPerChannel) => {
+        dailyStats.date = new Date(dailyStats.date);
+      });
```
sinergise-anze, line 203: Maybe it's just me, but with `forEach` I find it surprising that the original array changes. Maybe just a matter of taste, but I would suggest writing this a bit differently:
```javascript
for (let c in data) {
  data[c] = data[c].map((dailyStats: StatsPerChannel) => ({
    ...dailyStats,
    date: new Date(dailyStats.date);
  }));
}
```
(didn't run it)

src/layer/const.ts
```diff
@@ -99,3 +100,54 @@ export const SH_SERVICE_HOSTNAMES_V3: string[] = [
...
 ];
+
+export type GetStatsAndHistogramParams = {
+  fromTime: Date;
+  toTime: Date;
+  crs: CRS;
+  resolution: number;
+  geometry: Polygon;
+  binAmount?: number;
+  evalscript?: string;
```
sinergise-anze, line 111: I'm not convinced that this is the proper way - the "layer" is a separate visualization of data. If the evalscript changes, a new layer should be instantiated.

This concept allows you to leverage tools which are already there, for example `updateLayerFromServiceIfNeeded`. Once we implement `evalscriptUrl` fetching we can use that too.

In other words, I would suggest removing `evalscript` from params and using `this.evalscript` instead.

README.md
```diff
@@ -180,6 +180,29 @@ We can always use layer to search for data availability:
...
+```javascript
+  const layerS2L2A = new S2L2ALayer({
+    instanceId: '<my-instance-id>',
+    layerId: 'LAYER_S2L2A',
+    maxCloudCoverPercent: 100,
+  });
+
+const stats = await layerS2L2A.getStatsAndHistogram({
+  geometry: geometry, 
+  crs: CRS_EPSG4326,
```
sinergise-anze, line 197: We are a bit inconsistent here - with `BBox` we specify CRS as part of the structure, while here we are using a separate object. Not sure if it needs to be changed though (and how to best do it), what is your opinion?

guss84: The underlying FIS request requires a `crs` and the other reason is prefixing the resolution with `m`  when using `4326`, which would be nice to hide from the user in my view. So either we leave it as it is, or require a the crs property in the Geojson Geometry ie

```
type: "Polygon",
coordinates: [...]
crs:{
 type: "name",
 name: "urn:ogc:def:crs:EPSG::4326"
}
```

README.md
```diff
@@ -180,6 +180,29 @@ We can always use layer to search for data availability:
   const datesS1 = await layerS1.findDatesUTC(bbox, fromTime, toTime);
 ```
 
+## Getting basic statistics and histogram
+
+Getting basic statistics(mean, min, max, standard deviation) and a histogram for a geometry(Polygon or MultiPolygon).
```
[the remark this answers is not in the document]
guss84: Done

src/layer/AbstractLayer.ts
```diff
@@ -245,5 +245,10 @@ export class AbstractLayer {
     return await this.findDatesUTC(bbox, fromTime, toTime);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async getStatsAndHistogram(payload: any): Promise<any> {
```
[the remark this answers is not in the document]
guss84: Changed to `getStats`

stories/landsat5.eocloud.stories.js
guss84: Done

src/layer/const.ts
guss84: I made it uppercase, and commented it out

README.md
guss84: changed to `bins`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
guss84: Changed in `AbstractSentinelHubV1OrV2Layer.ts ` and `AbstractSentinelHubV3Layer.ts `

src/layer/AbstractSentinelHubV1OrV2Layer.ts
guss84: I just think it is a matter of taste, and I am indifferent between the 2. Changed it to use the for loop and map though

src/layer/const.ts
guss84: Sure, it is inconsistent with `getMap`, so changed to use `this.evalscript`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -140,4 +155,54 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+      throw new Error('Parameter "geometry" needs to be provided');
+    }
+    if (!params.crs) {
+      throw new Error('Parameter "crs" needs to be provided');
+    }
+    if (!params.resolution) {
+      throw new Error('Parameter "resolution" needs to be provided');
+    }
+    if (!params.fromTime || !params.toTime) {
+      throw new Error('Parameters "fromTime" and "toTime" needs to be provided');
```
sinergise-anze, line 170: Typo: need

(also below)

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -140,4 +155,54 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+      payload.evalscript = btoa(this.evalscript);
+      payload.evalsource = this.getEvalsource();
+    }
+
+    const { data } = await axios.get(this.dataset.shServiceHostname + 'v1/fis/' + this.instanceId, {
+      params: payload,
+    });
+    // convert date strings to Date objects
+    for (let channel in data) {
+      data[channel] = data[channel].map((dailyStats: StatsPerChannel) => ({
```
sinergise-anze, line 201: Sorry, I know I suggested this, but it's actually wrong. The input `dailyStats` is not of `StatsPerChannel` type (it has a string instead of a Date object). I would suggest:
```
data[channel] = data[channel].map((dailyStats: any) => ({
  ...
```

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -235,4 +250,51 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
+      ...this.getStatsAdditionalParameters(),
+    };
+    // When using CRS=EPSG:4326 one has to add the "m" suffix to enforce resolution in meters per pixel
+    if (params.crs.authId === CRS_EPSG4326.authId) {
+      payload.resolution = params.resolution + 'm';
+    } else {
+      payload.resolution = params.resolution;
+    }
+    if (this.evalscript) {
+      payload.evalscript = btoa(this.evalscript);
```
sinergise-anze, line 287: Nice, except - `btoa` is not defined in node.js. This should be:
```
  if (evalscript) {
    if (typeof window !== 'undefined' && window.btoa) {
      payload.evalscript = btoa(evalscript);
    } else {
      payload.evalscript = Buffer.from(evalscript, 'utf8').toString('base64');
    }
  }
```

Alternatively, we could define a new utility function (`agnosticBtoa` or just `btoa` - but then we need to make sure we use the correct one).

EDIT: the same above, in `AbstractSentinelHubV1OrV2Layer`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
guss84: True, but this makes me think that `StatsPerChannel` is wrongly named and would suggest `StatsPerDay`?

README.md
guss84: Yes we will need to expand `Polygon` as according to the spec the `crs` member was removed as can be seen at https://tools.ietf.org/html/rfc7946#appendix-B) under section `B.1.  Normative Changes` . Under section `4.  Coordinate Reference System` it is mentioned that all coordinates are in wgs:84/EPSG:4326. So yes we should default to `EPSG:4326`.

Do you still think it is a good idea to expand the `Polygon` declaration to include `crs`?

stories/landsat5.eocloud.stories.js
guss84: This one slipped through the big crack. Done

README.md
```diff
@@ -180,6 +180,29 @@ We can always use layer to search for data availability:
...
+    maxCloudCoverPercent: 100,
+  });
+
+const stats = await layerS2L2A.getStats({
+  geometry: geometry, 
+  crs: CRS_EPSG4326, 
+  fromTime: fromTime, 
+  toTime: toTime, 
+  resolution: resolution,
+  bins:10,
```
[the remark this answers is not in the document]
guss84: Fixed

src/layer/AbstractSentinelHubV3Layer.ts
guss84: True thanks, this has been fixed

src/layer/AbstractSentinelHubV1OrV2Layer.ts
sinergise-anze: Also true, yes... Though it's actually `StatsPerChannelPerDay`... :) Maybe `DailyChannelStats`?

README.md
sinergise-anze: I think it is a good idea to be consistent. :)

The problem here is that this is the interface and if we ever change it, it will be a breaking change, so we need to minimize the chance of this happening, or at least provide a way to upgrade interface in the future.

Since we probably don't want to change `BBox` definition, I would suggest adding `crs` to `Polygon`. If we ever decide that a better type is needed, we can introduce it and keep backwards-compatibility through object inspection. This way the decision shifts to "let's have a single parameter for shape and CRS", which sounds OK to me.

What do you think?

guss84: I removed `crs` as a param. The crs defaults to `EPSG:4326`, but you can specify the crs in the polygon's geometry with the following
```
type: "Polygon",
coordinates: [...]
crs:{
  type: "name",
  properties: {
    name: "urn:ogc:def:crs:EPSG::4326"
  }
}
```

Using our `Bbox` class also works by passing the `toGeojson()` for the `geometry` params as the method adds `crs` to the `Polygon`

guss84: The only issue is that `"urn:ogc:def:crs:OGC:1.3:CRS84"` would fail, since we assign `CRS_EPSG4326` to `CRS_WGS84`

src/layer/AbstractSentinelHubV1OrV2Layer.ts
guss84: Changed to `DailyChannelStats`

README.md
guss84: Changed the function to find the CRS from a `urn` to a switch case, and that resolves the issue where `urn:ogc:def:crs:OGC:1.3:CRS84` would throw an error

src/layer/const.ts
```diff
@@ -99,3 +100,52 @@ export const SH_SERVICE_HOSTNAMES_V3: string[] = [
...
+        lowEdge: number;
+        mean: number;
+        count: number;
+      },
+    ];
+  };
+};
+
+export type GetStats = {
+  [key: string]: [DailyChannelStats];
```
sinergise-anze, line 150: Just noticed: is this correct? IIUC, this is an array with a single element of `DailyChannelStats`? I think this is how it should be:
```
  [key: string]: DailyChannelStats[];
```

stories/landsat5.eocloud.stories.js
```diff
@@ -251,3 +251,33 @@ export const findDatesUTCEPSG3857 = () => {
...
+  const toTime = new Date(Date.UTC(2000, 3 - 1, 15, 23, 59, 59));
+
+  const perform = async () => {
+    const params = {
+      fromTime: fromTime,
+      toTime: toTime,
+      resolution: 100,
+      bins: 10,
+      geometry: bbox.toGeoJSON(),
+      crs: bbox.crs,
```
sinergise-anze, line 275: This line should be removed.

README.md
sinergise-anze: Nice, looks much better!

Artifact: this line (`crs: ...`) should be removed from README.

guss84: Thanks, removed the artifact in README and Landsat5 story

src/layer/const.ts
guss84: Yes you are correct, thanks for noticing

stories/landsat5.eocloud.stories.js
guss84: Removed

PR #64, opened 2020-04-08
Return most recent results first from findDatesUTC

sinergise-anze:
Service `findAvailableDates` is not consistent in the order in which the results are returned, so we check and reverse them if needed (we assume that the order is either `leastRecent` or `mostRecent`, not random).

### DIFF
--- a/src/layer/AbstractSentinelHubV1OrV2Layer.ts
+++ b/src/layer/AbstractSentinelHubV1OrV2Layer.ts
@@ -115,7 +115,7 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
     };
   }
 
-  protected getFindDatesUTCAdditionalParameters(): Record<string, any> {
+  protected async getFindDatesUTCAdditionalParameters(): Promise<Record<string, any>> {
     return {};
   }
 
@@ -128,7 +128,7 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
     const params = {
       timefrom: fromTime.toISOString(),
       timeto: toTime.toISOString(),
-      ...this.getFindDatesUTCAdditionalParameters(),
+      ...(await this.getFindDatesUTCAdditionalParameters()),
     };
 
     const url = `${this.dataset.findDatesUTCUrl}?${stringify(params, { sort: false })}`;
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -235,11 +235,9 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const response = await axios.post(this.dataset.findDatesUTCUrl, payload);
     const found: Moment[] = response.data.map((date: string) => moment.utc(date));
 
-    if (found.length > 1) {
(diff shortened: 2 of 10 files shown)


guss84:
I am pretty sure the order wont be random, but wouldnt it be safer to just sort the list and not make the assumption that the order is not random?

> Service `findAvailableDates` is not consistent in the order in which the results are returned, so we check and reverse them if needed (we assume that the order is either `leastRecent` or `mostRecent`, not random).

I am pretty sure the order wont be random, but wouldnt it be safer to just sort the list and not make the assumption that the order is not random?

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -233,6 +233,16 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       ...this.getFindDatesUTCAdditionalParameters(),
     };
     const response = await axios.post(this.dataset.findDatesUTCUrl, payload);
-    return response.data.map((date: string) => moment.utc(date).toDate());
+    const result = response.data.map((date: string) => moment.utc(date).toDate());
+
+    // S-5P, S-3 and possibly other datasets return the results in reverse order (leastRecent).
```
dgostencnik, line 238: This is also the case with byoc datasets. I'm not sure about sorting. It would probably be more clear and readable, but on the other hand in most of the cases it'll be just sorting already sorted array.

sinergise-anze: This was my assumption too, but @guss84 is right, I should check that this is true... I'll dig up this information and post it here.

sinergise-anze: It seems that this is a bug on the service side (I filed an issue), however it might take some time for this to be fixed. 

I have also noticed that `BYOCLayer` was re-implementing the same logic (but used `await` for `getFindDatesUTCAdditionalParameters`), so I fixed `getFindDatesUTCAdditionalParameters` to be `async` (no harm if function returns an object instead of Promise). 

I have also checked all other layers and added sorting to `WmsLayer`, so now `findDatesUTC` should always return sorted data, no matter which layer it is called on.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -115,7 +115,7 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
     };
   }
 
-  protected getFindDatesUTCAdditionalParameters(): Record<string, any> {
+  protected async getFindDatesUTCAdditionalParameters(): Promise<Record<string, any>> {
```
dgostencnik, line 118: Just curious why was this changed to async? fftr

sinergise-anze: Cleanup - `BYOCLayer` already defined it as `async` (because it needs to call `await this.fetchLayerParamsFromSHServiceV3`) and it's better to be consistent. 

This meant that `findDatesUTC` code was copy-pasted to `BYOCLayer` just to add an `await`... Since there is no harm in making this function `async`, I have simply changed the default signature and cleaned up duplicated code from `BYOCLayer`.

PR #65, opened 2020-04-08
evalscripturl for processingApi

dgostencnik:
This MR adds previously missing support for evalscriptUrl when using processingApi. If evalscripturl is set, script will be fetched from provided url and converted to version 3.

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -101,7 +101,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
         throw new Error('This layer does not support Processing API (unknown dataset)');
       }
       if (this.evalscriptUrl && !this.evalscript) {
-        const response = await axios.get(this.evalscriptUrl);
+        const response = await axios.get(this.evalscriptUrl, { responseType: 'text', useCache: true });
         const evalscriptV3 = await this.convertEvalscriptToV3(response.data);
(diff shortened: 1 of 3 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -100,8 +100,10 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       if (!this.dataset) {
         throw new Error('This layer does not support Processing API (unknown dataset)');
       }
-      if (this.evalscriptUrl) {
-        throw new Error('EvalscriptUrl is not supported with Processing API');
+      if (this.evalscriptUrl && !this.evalscript) {
+        const response = await axios.get(this.evalscriptUrl, { responseType: 'text', useCache: true });
+        const evalscriptV3 = await this.convertEvalscriptToV3(response.data);
```
sinergise-anze, line 105: Just a comment: this works great, I'm just not sure we should be hitting evalscript conversion service by default... I don't have a better suggestion though. I think adding a parameter would unnecessarily complicate interface, and the same with making this feature a separate function. And the request is cached.

What we could do is check the first line for `//VERSION=3` to avoid service calls at least in some cases - not sure it is worth the trouble though (most users are not very precise with these comments anyway, and since Processing API doesn't enforce them, there are many V3 scripts without the comments).

I think this is good to merge.

dgostencnik: Check for version was added as suggested.

PR #68, opened 2020-04-20
Fix S3SLSTR and L8@AWS stories

zcernigoj:
The stories for S3 SLSTR and L8@AWS had some problems (S3 SLSTR with the dates for findDatesUTC and L8@AWS with the layer for the stats).

This MR should fix those issues.


.env.example
```diff
@@ -18,6 +18,7 @@ BYOC_BBOX_EPSG3857_MAXX=
 BYOC_BBOX_EPSG3857_MAXY=
 S5PL2_LAYER_ID=
 LANDSAT8_LAYER_ID=
+LANDSAT8_NDVI_LAYER_ID=
```
dgostencnik, line 21: Don't forget to update .env file in out docs repo.

zcernigoj: Thanks, updated.

zcernigoj:
@sinergise-anze suggested that I should leave the story with the previous layer for NDVI for Landsat8 on AWS so it can be tested and investigated why it doesn't work anymore (it worked a few weeks ago).

I changed the layer back for the existing stories (with the exception that the layer ids are set in the .env file. I added the stories for the layer with evalscript V3, that works as of time of this MR.

stories/landsat8.aws.stories.js
```diff
@@ -278,10 +288,58 @@ export const stats = () => {
 export const statsBBOX3857 = () => {
```
sinergise-anze, line 288: Looks good!

I would just suggest renaming `stats` => `statsEvalscriptV2` and `statsBBOX3857` => `statsBBOX3857EvalscriptV2`, to make it more explicit what the point of these two stories is.

sinergise-anze: I have checked now why the old NDVI layer didn't work, by changing the story so that the `getMap` request with the same parameters was issued. It also didn't work... This was the original evalscript:

```

function evaluatePixel(sample) {
    let val = (sample.B05 - sample.B04) / (sample.B05 + sample.B04);
    return [val];
}

function setup(ds) {
  return {
    input: ["B04", "B05"],
    output: { bands: 1 }
  }
}
```

The problems:

- evalscript was using V3 syntax, not V2 (so the duplication of stories doesn't make sense - unless we with to have `V3` and `V3WithDataProduct`), but it wasn't marked as V3 (missing comment)
- `dataMask` was not used which caused FIS to break

I have now fixed the original NDVI layer because I don't think it ever worked in this form. My best guess is that someone converted it to V3 without checking, but without history we can't know.

I would suggest changing this back so that it only uses a single layer, preferably with `dataProduct`.

zcernigoj: Thanks for checking. It is becoming quite a task to check the evalscript nowadays :D 

Removed the duplicated stories and used only the layer with the `dataProduct`.

PR #70, opened 2020-04-22
fixed retries setting on axios interceptors

ignasieb:
Now retries can be passed as a config parameter on the request in order to specify the amount of retries that it should try.

### DIFF
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -206,8 +206,8 @@ const retryRequests = (err: any): any => {
   }
   if (shouldRetry(err.response.status)) {
     err.config.retriesCount = err.config.retriesCount | 0;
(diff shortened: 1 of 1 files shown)


src/utils/axiosInterceptors.ts
```diff
@@ -206,7 +206,8 @@ const retryRequests = (err: any): any => {
   }
   if (shouldRetry(err.response.status)) {
     err.config.retriesCount = err.config.retriesCount | 0;
-    const shouldRetry = err.config.retriesCount < RETRIES;
+    const minRetries = err.config.retries || RETRIES;
```
dgostencnik, line 209: This var should be probably named `maxRetries`

ignasieb: True, I already changed it. Thanks

PR #75, opened 2020-04-24
Feature/legend info

Ardweaden:
Adds `legendUrl` to all layers and `legend` to `SentinelHubV3Layer`. 

`legend` and `legendUrl` for `SentinelHubV1OrV2Layer` are fetched on `updateLayerFromServiceIfNeeded`.


zcernigoj:
In the layers' constructors, there are 2 ways you defined the `legendUrl` (everything regarding it is the same everywhere (`ConstructorParameters` , it's present as a parameter in a `constructor` call) ):

- `super({ ..., legendUrl });`
  - `AbstractSentinelHubV1OrV2Layer.ts`
  - `AbstractSentinelHubV3Layer.ts` (and `AbstractSentinelHubV3WithCCLayer.ts`, `S1GRDAWSEULayer.ts`, `S3SLSTRLayer.ts`, `5PL2Layer.ts`, )
  

- `this.legendUrl = legendUrl;`
  - `WmsLayer.ts`

Any reason, why it isn't set like `super({ ..., legendUrl });` in `WmsLayer.ts` also? (it extends the `AbstractLayer.ts` so it could be the same as elsewhere)

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -210,4 +213,26 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
     }
     return data;
   }
+
+  public async updateLayerFromServiceIfNeeded(): Promise<void> {
+    if (this.instanceId === null || this.layerId === null) {
+      throw new Error(
+        "One or more of these parameters (polarization, acquisitionMode, resolution) \
```
zcernigoj, line 220: Not every dataset on EOCloud has `polarization, acquisitionMode, resolution` parameters, so I think this message should be changed (maybe to `Some or more dataset-specific parameters are not set ...`).

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -330,4 +337,9 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const res = await axios.post(url, evalscript, requestConfig);
     return res.data;
   }
+
+  public async updateLayerFromServiceIfNeeded(): Promise<void> {
```
zcernigoj, line 341: `S1GRDAWSEULayer.ts` extends `AbstractSentinelHubV3Layer` and already has `updateLayerFromServiceIfNeeded()` so it would override it.

src/layer/LayersFactory.ts
```diff
@@ -194,18 +196,26 @@ export class LayersFactory {
     filterLayers: Function | null,
   ): Promise<AbstractLayer[]> {
     const parsedXml = await fetchGetCapabilitiesXml(baseUrl);
+
     const layersInfos = parsedXml.WMS_Capabilities.Capability[0].Layer[0].Layer.map(layerInfo => ({
       layerId: layerInfo.Name[0],
       title: layerInfo.Title[0],
       description: layerInfo.Abstract ? layerInfo.Abstract[0] : null,
       dataset: null,
+      legendUrl:
```
zcernigoj, line 205: Could this also be done in the `WmsLayer.ts`, similarly to `AbstractSentinelHubV1OrV2Layer.ts`? (in case you want to create a WMS layer manually and you want to have the legend)

Ardweaden:
> In the layers' constructors, there are 2 ways you defined the `legendUrl` (everything regarding it is the same everywhere (`ConstructorParameters` , it's present as a parameter in a `constructor` call) ):
> 
> * `super({ ..., legendUrl });`
>   
>   * `AbstractSentinelHubV1OrV2Layer.ts`
>   * `AbstractSentinelHubV3Layer.ts` (and `AbstractSentinelHubV3WithCCLayer.ts`, `S1GRDAWSEULayer.ts`, `S3SLSTRLayer.ts`, `5PL2Layer.ts`, )
> * `this.legendUrl = legendUrl;`
>   
>   * `WmsLayer.ts`
> 
> Any reason, why it isn't set like `super({ ..., legendUrl });` in `WmsLayer.ts` also? (it extends the `AbstractLayer.ts` so it could be the same as elsewhere)

There is no reason... Thanks, fixed!

Btw, I'm sorry for responding so late. I thought no one has taken the CR and didn't check back in a while. 
@zcernigoj (how the hell does one reply to comments on Github)

src/layer/AbstractSentinelHubV1OrV2Layer.ts
Ardweaden: Very true, thanks for noticing! I changed it to `Additional data can't be fetched from service because instanceId and layerId are not defined`. Is it fine with you?

The other comment "assumed" the update was requested because those parameters were not set and thus the layer needed to fetch that additional data. In our case, this is not the premise - in my opion.

src/layer/AbstractSentinelHubV3Layer.ts
Ardweaden: Should be fixed.

src/layer/LayersFactory.ts
Ardweaden: I'm sorry, but I don't quite understand what you mean.

Ardweaden:
> Code review in the comments above.
> 
> Also, please provide some example (storybook / node example) if you have time.

@zcernigoj  
This is being used here: https://git.sinergise.com/team-6/eobrowser3/merge_requests/108 and can be tested in EOB3 directly.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
zcernigoj: Great, sounds better than my suggestion. Resolving.

src/layer/AbstractSentinelHubV3Layer.ts
zcernigoj: Resolving.

src/layer/LayersFactory.ts
zcernigoj: in `AbstractSentinelHubV1OrV2Layer.ts` you added the function `updateLayerFromServiceIfNeeded` with the code that fetches the legend url if you (manually) create a layer that extends `AbstractSentinelHubV1OrV2Layer` without the legend url.

Currently you can manually create a `WmsLayer` without legend url but there is no way to update the `WmsLayer` with additional properties, for example the legend url (no `updateLayerFromServiceIfNeeded`).

The code of `updateLayerFromServiceIfNeeded` in `WmsLayer` would be the same as in `AbstractSentinelHubV1OrV2Layer` because the OGC `GetCapabilities` is in XML format.

zcernigoj:
> There is no reason... Thanks, fixed!

Nice

> @zcernigoj (how the hell does one reply to comments on Github)

I don't think there is a way to reply to a comment (that is not a comment on the code).

src/layer/LayersFactory.ts
Ardweaden: You are correct. 

However, - correct me if I'm wrong - I don't think you can update any other info for any other layer either, except for those specific parameters in `updateLayerFromServiceIfNeeded`. For example, if one creates a layer, I don't think they can get the `description` from service? I might be wrong. This should probably be implemented?

The slight issue I have with this is, that `updateLayerFromServiceIfNeeded` generally fetches data from a different endpoint that `makeLayers` use. However in this case, it would be exactly the same. So `makeLayer(instanceId, layerId).updateLayerFromServiceIfNeeded()` will basically do the same thing twice for a `WMSLayer`. 

Additionally, I don't think we should override the manually set values.

Therefore, I decided *not* to update if `legendUrl` is already defined.

zcernigoj: > However, - correct me if I'm wrong - I don't think you can update any other info for any other layer either, except for those specific parameters in updateLayerFromServiceIfNeeded. 

You are correct. 

> For example, if one creates a layer, I don't think they can get the description from service? I might be wrong. This should probably be implemented?

I think they can get the description from the service in any other way than requesting `getCapabilities`. For Sentinel Hub layers, it's only possible if you have the instance id ([docs](https://docs.sentinel-hub.com/api/latest/reference/#operation/getWmsLayers)). I don't think we need getting the layer description at this moment.

> Therefore, I decided not to update if legendUrl is already defined.

:+1: I think it's great like that.

PR #76, opened 2020-04-29
Fix/convert-evalscripts-v3

ignasieb:
Now sentinelhub-js converts evalscripts to V3 when the layer contains the `evalscripturl` (already implemented) and when it has the evalscript stored in `this.evalscript` (this MR).
This conversion would only happen when the processing API is being used.
Any changes that you feel that are necessary, please let me know.

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -35,6 +35,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   protected evalscript: string | null;
   protected evalscriptUrl: string | null;
   protected dataProduct: string | null;
+  protected evalscriptWasConvertedToV3: boolean | null;
 
   public constructor({
     instanceId = null,
@@ -61,6 +62,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     this.evalscript = evalscript;
     this.evalscriptUrl = evalscriptUrl;
     this.dataProduct = dataProduct;
+    this.evalscriptWasConvertedToV3 = false;
   }
 
   protected async fetchLayerParamsFromSHServiceV3(): Promise<any> {
@@ -112,14 +114,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       }
       if (this.evalscriptUrl && !this.evalscript) {
         const response = await axios.get(this.evalscriptUrl, { responseType: 'text', useCache: true });
-        let evalscriptV3;
-        //Check version of fetched evalscript by checking if first line starts with //VERSION=3
-        if (response.data.startsWith('//VERSION=3')) {
-          evalscriptV3 = response.data;
-        } else {
-          evalscriptV3 = await this.convertEvalscriptToV3(response.data);
-        }
-        this.evalscript = evalscriptV3;
+        this.evalscript = response.data;
       }
       if (!this.evalscript && !this.dataProduct) {
         const layerParams = await this.fetchLayerParamsFromSHServiceV3();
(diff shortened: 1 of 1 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -132,6 +132,12 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
         }
       }
 
+      //Convert internal evalscript to V3 if it's not in that version.
+      if (this.evalscript && !this.evalscript.startsWith('//VERSION=3')) {
+        let evalscriptV3 = await this.convertEvalscriptToV3(this.evalscript);
+        this.evalscript = evalscriptV3;
+      }
+
```
sinergise-anze, line 140: Minor cleanup: since we are now converting to V3 here, can we maybe get rid of the lines 116-121?

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -132,6 +132,12 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
         }
       }
 
+      //Convert internal evalscript to V3 if it's not in that version.
+      if (this.evalscript && !this.evalscript.startsWith('//VERSION=3')) {
+        let evalscriptV3 = await this.convertEvalscriptToV3(this.evalscript);
```
sinergise-anze, line 137: We are making an assumption here that the service will return an evalscript which starts with `'//VERSION=3'`. While this is true now, it is an undocumented feature. 

I know we did the same thing before and that this is not something you have changed, but I would suggest changing `convertEvalscriptToV3()` so that when evalscript is fetched, this condition is checked, and if false, the comment is prepended... Just to be on the safe side. If you feel this is out of scope feel free to resolve and merge.

Ardweaden: I am not sure whether this handled somehow, but what about basic evalscripts? Afaik evalscript converter *does not* convert basic evalscripts to full V3 format, and prepending `//VERSION=3` would result in an error on rendering.

src/layer/AbstractSentinelHubV3Layer.ts
ignasieb: Lines 116-121 convert evalscripts given its url. So I think that they shouldn't be deleted.

sinergise-anze: I meant, we should replace lines 115-122 with:
```
this.evalscript = response.data;
```

Then the code from line 135 on will take care of conversion (if needed). Or am I missing something?

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: Good point! But this also means that the check `this.evalscript.startsWith('//VERSION=3')` is incomplete, since we are converting basic scripts every time, if they were already converted or not.  Not sure how to fix this though... Do we keep the original evalscript and put the converted one to `this.evalscriptV3`? Or do we set `this.evalscriptWasConvertedToV3` to true?

Ardweaden: I think it's almost certain that a script is basic of it doesn't contain `function evaluatePixel`, `function setup`, `setInputComponents(` etc. which would be possible to check, right? They have to be reserved names. After all, this is what service must be doing already.

src/layer/AbstractSentinelHubV3Layer.ts
ignasieb: Yes of course, sorry I didn't get you.

src/layer/AbstractSentinelHubV3Layer.ts
ignasieb: For what I'm checking, basic scripts don't get converted to V3 and they don't get prepended the //VERSION=3 comment.
Should we check if a script is basic in order to avoid doing unnecessary requests to the convertscript endpoint?

sinergise-anze: I think it will be easier (and less error-prone) if we simply set `this.evalscriptWasConvertedToV3 = true;`) and then skip conversion if this flag is set. What do you think?

ignasieb: Yes of course. I will do that. Thanks Anze.

ignasieb: Can I merge it as it is right now Anze? @sinergise-anze

sinergise-anze: Yes, absolutely! Looks good.

PR #78, opened 2020-04-30
Feature/cancelling requests

ignasieb:
Now it's possible to cancel and retry requests a certain amount of times by passing a configuration object (type: AxiosRequestConfig) to the getMap (both WMS and PROCESSING) and getStats.
There are new exports on index.ts to fulfill this functionality, one is cancelFactory that creates a factory of tokens and the other is isCancelled that allows to check the expection thrown by cancelling a request.
To test this there's a new example on `example/node/index.js` and a new story on S2L1C stories.

Anything you feel that needs to be changed, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,18 +193,18 @@ We can always use layer to search for data availability:
 
 You can also cancel requests when searching/fetching data.
 
-To do so a factory of tokens needs to be created and a token passed through a request configuration object. In this object can also be defined the amount of retires desired in the case the petition fails.
+To do so a token needs to be created and passed through a requests configuration object. Other config such as retries can also be defined there.
 
 In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
 
-This exception can be caught and identified by isCancelled, function that can be imported from the library.
+This exception can be caught and identified by `isCancelled`.
 
 ```typescript
-import { cancelFactory, isCancelled, RequestConfiguration } from '@sentinel-hub/sentinelhub-js';
+import { CancelFactory, isCancelled } from '@sentinel-hub/sentinelhub-js';
 
 const source = CancelFactory.createSource();
 
-const requestConfig : RequestConfiguration = {
+const requestConfig = {
   cancelToken: source.getToken(),
   retries: 4
 }
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ try {
   const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestConfig);
   const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
   const stats = await layer.getStats(getStatsParams, requestConfig);
-  const tiles = await layer.findTiles(bbox, fromTime, toTime, undefined, undefined, requestConfig);
+  const tiles = await layer.findTiles(bbox, fromTime, toTime, null, null, requestConfig);
 }
 catch(err) {
   if (!isCancelled(err)) {
--- a/README.md
+++ b/README.md
@@ -188,6 +188,42 @@ We can always use layer to search for data availability:
   const datesS1 = await layerS1.findDatesUTC(bbox, fromTime, toTime);
 ```
 
+
+## Cancelling requests
+
+You can also cancel requests when searching/fetching data.
+
+To do so a factory of tokens needs to be created and a token passed through a request configuration object. In this object can also be defined the amount of retires desired in the case the petition fails.
+
+In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
+
+This exception can be caught and identified by isCancelled, function that can be imported from the library.
+
+```typescript
+import { cancelFactory, isCancelled, RequestConfiguration } from '@sentinel-hub/sentinelhub-js';
+
+const source = cancelFactory();
+
+const requestConfig : RequestConfiguration = {
+  cancelToken: source.token,
+  retries: 4
+}
+
+setTimeout(() => {
+  source.cancel();
+}, 500);
+
+try {
+  const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestConfig);
+  const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
+}
+catch(err) {
+  if (isCancelled(err)) {
+    // Exception thrown by the cancelled requests is going to be identified by isCancelled.
+  }
+}
+````
+
 ## Getting basic statistics and histogram
 
 Getting basic statistics (mean, min, max, standard deviation) and a histogram for a geometry (Polygon or MultiPolygon).
--- a/README.md
+++ b/README.md
@@ -202,10 +202,10 @@ This exception can be caught and identified by isCancelled, function that can be
 ```typescript
 import { cancelFactory, isCancelled, RequestConfiguration } from '@sentinel-hub/sentinelhub-js';
 
-const source = cancelFactory();
+const source = CancelFactory.createSource();
 
 const requestConfig : RequestConfiguration = {
-  cancelToken: source.token,
+  cancelToken: source.getToken(),
   retries: 4
 }
 
@@ -216,10 +216,11 @@ setTimeout(() => {
 try {
   const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestConfig);
   const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
+  const stats = await layer.getStats(getStatsParams, requestConfig);
 }
 catch(err) {
-  if (isCancelled(err)) {
-    // Exception thrown by the cancelled requests is going to be identified by isCancelled.
+  if (!isCancelled(err)) {
+    throw err;
   }
 }
 ````
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ try {
   const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestConfig);
   const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
   const stats = await layer.getStats(getStatsParams, requestConfig);
+  const tiles = await layer.findTiles(bbox, fromTime, toTime, undefined, undefined, requestConfig);
 }
 catch(err) {
   if (!isCancelled(err)) {
--- a/README.md
+++ b/README.md
@@ -195,22 +195,22 @@ You can also cancel requests when searching/fetching data.
 
 To do so a token needs to be created and passed through a requests configuration object. Other config such as retries can also be defined there.
 
-In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
+In the example below, a cancel token is passed inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
 
 This exception can be caught and identified by `isCancelled`.
 
 ```typescript
-import { CancelFactory, isCancelled } from '@sentinel-hub/sentinelhub-js';
+import { CancelToken, isCancelled } from '@sentinel-hub/sentinelhub-js';
 
-const source = CancelFactory.createSource();
+const token = new CancelToken();
 
 const requestConfig = {
-  cancelToken: source.getToken(),
+  cancelToken: token,
   retries: 4
 }
 
 setTimeout(() => {
-  source.cancel();
+  token.cancel();
 }, 500);
 
 try {
--- a/example/node/index.js
+++ b/example/node/index.js
@@ -13,7 +13,7 @@ const {
   MimeTypes,
   ApiType,
   setDebugEnabled,
-  CancelFactory,
+  CancelToken,
   isCancelled,
 } = require('../../dist/sentinelHub.cjs');
 
@@ -190,13 +190,15 @@ async function run() {
 
   // Cancel request example
   // Create source and pass token as config on the request (getMap WMS/Processing and getStats supported for layerV3)
-  const source = CancelFactory.createSource();
+  const token = new CancelToken();
+
   setTimeout(() => {
     printOut('Cancelling request');
-    source.cancel();
+    token.cancel();
   }, 100);
+
   try {
-    const response = await layerS2L2A.getMap(getMapParams, ApiType.WMS, { cancelToken: source.token });
+    const response = await layerS2L2A.getMap(getMapParams, ApiType.WMS, { cancelToken: token });
     console.log(response);
   } catch (err) {
     //The exception thrown by cancelling requests can be identified by isCancelled
--- a/example/node/index.js
+++ b/example/node/index.js
@@ -196,9 +196,7 @@ async function run() {
     source.cancel();
   }, 100);
   try {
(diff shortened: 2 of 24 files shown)


sinergise-anze:
First of all, thank you for taking this task, it is much needed! (and we've been putting it off for way too long...)

There are a few things that I would suggest solving a bit differently:
- we shouldn't expose axios to the outside world (we need to "cast" axios types to our own) as this is just our internal implementation detail, and it might even change at some point in the future.
- I would suggest we don't add `requestConfig` to `getMap` because user should not care about the request and how it is done. We can either add another parameter (`cancelToken`), or we can put both `apiType` and `cancelToken` parameters inside the `GetMapParams` (backwards incompatible change) - don't have a strong opinion here though
- `getMap` is only one of the functions which should accept a cancel token; ideally, every async function should be able to do it. Not sure how to implement this though - should we find a way to trigger this globally, or for all functions on a layer?

Before changing anything, maybe it would be easier to describe a solution in README? This way we can discuss it before spending more time on implementation.

ignasieb:
After this commits, all Axios requests should be covered by the possibility of canceling requests using a RequestConfiguration object where you can specify the cancellation token and retries.
This MR changes a lot of files because of this. 
Readme.md has a description of how the cancellation should be done, and there's also an example on index.js and a story on S2L1C stories.

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
   const datesS1 = await layerS1.findDatesUTC(bbox, fromTime, toTime);
 ```
 
+
+## Cancelling requests
+
+You can also cancel requests when searching/fetching data.
+
+To do so a factory of tokens needs to be created and a token passed through a request configuration object. In this object can also be defined the amount of retires desired in the case the petition fails.
```
sinergise-anze, line 196: - > ...a factory of tokens needs to be created...

  Technically, factory is not created, it is just used to create a token.

- Typo: `retires`

- `request configuration object` -> `requests configuration object` (it configures multiple requests)

I would suggest:

```
To do so a token needs to be created and passed through a requests configuration object. Other config such as retries can also be defined there.
```

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
...
+
+In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
+
+This exception can be caught and identified by isCancelled, function that can be imported from the library.
```
sinergise-anze, line 200: - isCancelled => `isCancelled` (backticks should be added)

- I would suggest leaving out that `isCancelled` can be imported from the library (it is true for everything else too).

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
...
+
+```typescript
+import { cancelFactory, isCancelled, RequestConfiguration } from '@sentinel-hub/sentinelhub-js';
```
sinergise-anze, line 203: Typo: `cancelFactory` -> `CancelFactory`

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
...
+
+const source = CancelFactory.createSource();
+
+const requestConfig : RequestConfiguration = {
```
sinergise-anze, line 207: We are using JavaScript (and not TypeScript) in the README, so I would suggest removing `RequestConfiguration` from here and from the import.

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
...
+To do so a factory of tokens needs to be created and a token passed through a request configuration object. In this object can also be defined the amount of retires desired in the case the petition fails.
+
+In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
+
+This exception can be caught and identified by isCancelled, function that can be imported from the library.
+
+```typescript
+import { cancelFactory, isCancelled, RequestConfiguration } from '@sentinel-hub/sentinelhub-js';
+
+const source = CancelFactory.createSource();
```
sinergise-anze, line 205: I must confess that I never understood why axios distinguishes between "source" and "token"... Do you think we could do something like this?
```
const cancelToken = CancelFactory.createToken();
...
const requestConfig : RequestConfiguration = {
  cancelToken: cancelToken,
  retries: 4, // request is tried up to 5 times
}
```

example/node/index.js
```diff
@@ -185,6 +187,25 @@ async function run() {
...
+  // Create source and pass token as config on the request (getMap WMS/Processing and getStats supported for layerV3)
+  const source = CancelFactory.createSource();
+  setTimeout(() => {
+    printOut('Cancelling request');
+    source.cancel();
+  }, 100);
+  try {
+    const response = await layerS2L2A.getMap(getMapParams, ApiType.WMS, { cancelToken: source.getToken() });
+    // const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
+    // const stats = await layer.getStats(getStatsParams, requestConfig);
```
sinergise-anze, line 201: Is this something that should be cleaned up?

src/index.ts
```diff
@@ -108,6 +109,8 @@ export {
   BBox,
   LocationIdSHv3,
   setDebugEnabled,
+  CancelFactory,
+  isCancelled,
```
sinergise-anze, line 113: We are probably missing `RequestConfiguration` export here?

src/layer/AbstractLayer.ts
```diff
@@ -23,14 +29,15 @@ export class AbstractLayer {
...
-  public async getMap(params: GetMapParams, api: ApiType): Promise<Blob> {
+  public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
     switch (api) {
       case ApiType.WMS:
         const url = this.getMapUrl(params, api);
         const requestConfig: AxiosRequestConfig = {
           // 'blob' responseType does not work with Node.js:
           responseType: typeof window !== 'undefined' && window.Blob ? 'blob' : 'arraybuffer',
           useCache: true,
+          ...reqConfig,
```
sinergise-anze, line 40: This is an elegant solution, the only problem is that users might find this a shortcut so they can provide the parameters directly to the axios. Not sure how to best solve this though... Specifying the parameters one by one also doesn't sound nice. Do you have some other idea? If not, fftr.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -245,12 +263,19 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
   ): Promise<{ data: { tiles: any[]; hasMore: boolean } }> {
     if (!searchIndexUrl) {
       throw new Error('This dataset does not support searching for tiles');
     }
+    if (!maxCount) {
+      maxCount = 1;
+    }
+    if (!offset) {
+      offset = 0;
+    }
```
sinergise-anze, line 278: To make it clearer what is happening, I would suggest changing the default values to `null` and checking explicitly for `null`, not just falsey:
```
    maxCount: number | null = null,
    offset: number | null = null,
...
    if (maxCount === null) {
      maxCount = 1;
    }
    if (offset === null) {
      offset = 0;
    }
```

src/utils/cancelRequests.ts
```diff
@@ -0,0 +1,24 @@
+import axios, { CancelTokenSource, CancelToken } from 'axios';
+
+export class CancelFactory {
+  protected source: CancelTokenSource;
+  private constructor() {
+    this.source = axios.CancelToken.source();
+  }
+
+  public static createSource = (): CancelFactory => {
+    return new CancelFactory();
```
sinergise-anze, line 10: With factory patter, you are not creating the object of `CancelFactory` - you should be creating the target object, in this case `CancelTokenSource` or `CancelToken` (depends on the interface you decide to implement). 

In ~~other words~~ code:
```
export class CancelFactory {
  protected static source: cancelTokenSource | null = null;

  public static createSource(): CancelTokenSource {
    if (CancelFactory.cancelTokenSource === null) {
      CancelFactory.cancelTokenSource = axios.CancelToken.source();
    }
    return CancelFactory.cancelTokenSource;
  }
}
```

Note that you don't need the `getToken` and `cancel` methods because they should be called on returned value (source) anyway.

example/node/index.js
ignasieb: The idea was to put examples on the different methods that can be cancelled and a visual example on how to request configuration was passed on. I agree it's not the cleanest thing, and maybe it's better to just leave the getMap on it.

src/layer/AbstractLayer.ts
ignasieb: The idea is that because RequestConfiguration has a type, users cannot shorcut parametrs of axios request into RequestConfiguration, since this type only allows for certain key-value pairs, such as useCache, retries and cancelToken (if other parameters are specified an exception should be thrown?)
Maybe this doesn't prevent the problem the way I thought so some modifications are needed.

src/layer/AbstractSentinelHubV3Layer.ts
ignasieb: Sure, that's better, since passing undefined does not affect the default values.

README.md
ignasieb: I don't think so, since the source is the object needed to cancel the requests later on.
With the token alone you can't do that. It's not as convenient but I don't see any way to do it.

src/index.ts
ignasieb: True, thanks!

example/node/index.js
sinergise-anze: Not necessarily, you can also just uncomment them... IIUC the example works this way too? (of course, `requestConfig` needs to be defined or replaced)

src/layer/AbstractLayer.ts
sinergise-anze: Yes, but this only works if developers are using TypeScript, JavaScript will allow anything to be passed. But as I said, I don't see another clean solution, so fftr.

README.md
sinergise-anze: Sorry if I wasn't clear... It is up to us to return whatever we want as "token". We can easily return "axios source" and then use `getToken()` on it when we need the token, internally - this is up to us. Or we could return an object which holds the source _and_ the token, if needed.

My concern here is that we might be leaking too much of axios implementation details to the outside.

README.md
```diff
@@ -188,6 +188,44 @@ We can always use layer to search for data availability:
...
 ```
 
+
+## Cancelling requests
+
+You can also cancel requests when searching/fetching data.
+
+To do so a token needs to be created and passed through a requests configuration object. Other config such as retries can also be defined there.
+
+In the example below, a cancelFactory is created, passing a token inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
```
sinergise-anze, line 198: Sorry, missed this:
```
In the example below, a cancel token is passed inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
```

(we are not creating a `CancelFactory` object)

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: This is not what I meant, sorry if I wasn't clear. I don't think the example in README works right now. I think these changes are needed:
- example in README should use `null` instead of `undefined` for `findTiles` parameters `maxCount` and `offset`
- definition of `maxCount` and `offset` parameters should be changed as above (`null` is allowed and is a default value) - in **every** definition of `findTiles`, and especially in `AbstractLayer`:
  ```
  maxCount: number | null = null,
  offset: number | null = null,
  ```
- default value for `maxCount` needs to be consistent; since it was `50` before (see `AbstractLayer`), I would suggest we keep `50`: 
  ```javascript
  if (maxCount === null) {
    maxCount = 50;
  }
  ```

example/node/index.js
```diff
@@ -185,6 +187,25 @@ async function run() {
...
+  const token = new CancelToken();
+
+  setTimeout(() => {
+    printOut('Cancelling request');
+    token.cancel();
+  }, 100);
+
+  try {
+    const response = await layerS2L2A.getMap(getMapParams, ApiType.WMS, { cancelToken: token });
+    console.log(response);
```
sinergise-anze, line 202: Nitpick: since we don't usually output an image to the console, I would suggest:
```
console.log("Image was received in less than 100ms.");
```

Fftr.

src/layer/S3OLCILayer.ts
```diff
@@ -1,7 +1,9 @@
 import moment from 'moment';
 
 import { BBox } from 'src/bbox';
-import { PaginatedTiles, Link, LinkType } from 'src/layer/const';
+
+import { PaginatedTiles, Link, LinkType, RequestConfiguration } from 'src/layer/const';
+
```
sinergise-anze, line 6: Very very very very minor nitpick: newlines are tipically used to distinguish between system imports (from external packages) and local imports, so it makes sense to remove them here. It's more a styling rule though, fftr either way.

EDIT: same below in S3SLSTR and S5P.

src/utils/cancelRequests.ts
```diff
@@ -0,0 +1,35 @@
+import axios, { CancelTokenSource, AxiosRequestConfig, CancelToken as CancelTokenAxios } from 'axios';
+import { RequestConfiguration } from '../layer/const';
```
sinergise-anze, line 2: Not sure why `npm run build` doesn't warn about it, but there is a circular import here (`src/utils/cancelRequests.ts` => `src/layer/const.ts` => `src/utils/cancelRequests.ts`). I think it doesn't pose any problems in this case, but I would still suggest moving `RequestConfiguration` to this file - it belongs here a bit more anyway (imho).

Nitpick: by convention we use absolute imports in this library (`from 'src/layer/const';` instead of `from '../layer/const'`).

example/node/index.js
sinergise-anze: Type: `recieved` should be `received`. Fftr.

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -109,12 +112,19 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+    reqConfig?: RequestConfiguration,
   ): Promise<PaginatedTiles> {
     if (!this.dataset.searchIndexUrl) {
       throw new Error('This dataset does not support searching for tiles');
     }
+    if (maxCount === null) {
+      maxCount = DEFAULT_FIND_TILES_MAX_COUNT_PARAMETER;
+    }
+    if (offset === null) {
+      offset = DEFAULT_FIND_TILES_OFFSET_PARAMETER;
```
sinergise-anze, line 126: Minor nitpick: `DEFAULT_FIND_TILES_OFFSET_PARAMETER` is not needed, it is clearer if it is set to `0` directly (because any other value of the constant doesn't make sense). Fftr.

src/layer/AbstractLayer.ts
```diff
@@ -75,6 +78,7 @@ export class AbstractLayer {
     toTime: Date,
     maxFindTilesRequests: number = 50,
     tilesPerRequest: number = 50,
+    reqConfig?: RequestConfiguration,
```
sinergise-anze, line 81: We have missed `findFlyovers` from the README, and the default values are `50` & `50` instead of `null` (as with `findTiles`). But we can leave this for a future cleanup if you wish. Fftr.

PR #79, opened 2020-05-01
Fix/legacy proper layers

sinergise-anze:
This MR changes behaviour of `legacyGetMapFromParams` so that it uses layers from `LayersFactory` instead of always using a `WmsLayer`. The reason for this change is EO Browser, where we wish to work with BYOC layers which (even though their canonical WMS base URL points at eu-central-1 data location) need to be fetched from other locations.

There is also a breaking change to `parseLegacyWmsGetMapParams`, which now returns evalscript in decoded form (it is no longer base64 encoded). Since it is not used anywhere yet, it should be safe to do it now.

### DIFF
--- a/stories/legacy.stories.js
+++ b/stories/legacy.stories.js
@@ -1,6 +1,6 @@
 import { setAuthTokenWithOAuthCredentials } from './storiesUtils';
 
-import { legacyGetMapFromUrl, ApiType, BBox, CRS_EPSG3857, legacyGetMapFromParams } from '../dist/sentinelHub.esm';
+import { legacyGetMapFromUrl, ApiType, legacyGetMapFromParams } from '../dist/sentinelHub.esm';
 
 if (!process.env.INSTANCE_ID) {
   throw new Error('INSTANCE_ID environment variable is not defined!');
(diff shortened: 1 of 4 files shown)


stories/legacy.stories.js
```diff
@@ -1,6 +1,6 @@
 import { setAuthTokenWithOAuthCredentials } from './storiesUtils';
 
-import { legacyGetMapFromUrl, ApiType } from '../dist/sentinelHub.esm';
+import { legacyGetMapFromUrl, ApiType, BBox, CRS_EPSG3857, legacyGetMapFromParams } from '../dist/sentinelHub.esm';
```
zcernigoj, line 3: `BBox, CRS_EPSG3857` seem to not be used in here.

zcernigoj:
`legacyGetMapFromParams` is used in Playground also, so I tested if everything still works and it seems everything is ok. 👍

stories/legacy.stories.js
sinergise-anze: Thanks, fixed.

zcernigoj: Resolving.

PR #80, opened 2020-05-05
fix locationId null error for getStats

guss84:
* Fix issue where calling `getStats` on `BYOCLayer` would throw an error `'Parameter locationId must be specified'`

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -11,7 +11,7 @@ import {
   GetMapParams,
   ApiType,
   GetStatsParams,
-  GetStats,
+  Stats,
 } from 'src/layer/const';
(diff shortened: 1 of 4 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -10,6 +10,8 @@ import {
   SHV3_LOCATIONS_ROOT_URL,
   GetMapParams,
   ApiType,
+  GetStatsParams,
+  GetStats,
```
sinergise-anze, line 14: I never notice it before... shouldn't this be named just `Stats`? :)

Fftr & merge!

guss84: Changed the type name to `Stats`

PR #81, opened 2020-05-06
Do not break if evalsource is wrong, just warn

sinergise-anze:
Instead of being strict with `evalsource` parameter, we just warn when it is invalid (fixes EOB bug: custom scripts no longer working on user instances).


src/legacyCompat.ts
```diff
@@ -50,7 +50,7 @@ export async function legacyGetMapFromParams(
     // that doesn't match layer's dataset - but we check it nevertheless:
     const expectedEvalsource = layer.dataset.shWmsEvalsource;
     if (expectedEvalsource !== evalsource) {
-      throw new Error(
+      console.warn(
         `Evalsource ${evalsource} is not valid on this layer (was expecting ${expectedEvalsource})`,
```
guss84, line 54: I would change the message a bit to notify users that we will be using `expectedEvalsource`.

FFTR and merge

sinergise-anze: Thanks, did, merging.

PR #82, opened 2020-05-07
Feature/map data manipulation

zcernigoj:
First version of functions for map data manipulation. Currently only `gain` and `gamma` parameters are supported (as a part of `GetMapParams` ... Question: Should they be separate params for `getMap()` or part of maybe `manipulationParams` or something?).

If you send the `gain` and/or `gamma` parameters (later also other parameters for more advanced image manipulation) to `getMap()` (doesn't matter if API type is `processing` or `WMS`), first the image with no effects is retrieved from the services and then the functions are used to change the image.

---

`src/mapDataManipulation` folder contains these files:
- `src/mapDataManipulation/mapDataManipulation.ts` contains code that changes the pixels' values
- `src/mapDataManipulation/const.ts` contains constants (similar to `src/layer/const.ts`)
- `src/mapDataManipulation/mapDataManipulationUtils.ts` contains util functions.    
- `src/mapDataManipulation/predefinedEffectFunctions.ts` contains predefined functions (currently just gain, gamma)
- `src/mapDataManipulation/runPredefinedEffectFunctions.ts` contains a function that uses functions from `src/mapDataManipulation/predefinedEffectFunctions.ts`.    

Function from `src/mapDataManipulation/runPredefinedEffectFunctions.ts` is used in:
- `AbstractLayer.ts` (`AbstractSentinelHubV1OrV2Layer.ts` uses `getMap()` from `AbstractLayer.ts`)
- `AbstractSentinelHubV3Layer.ts`

---

Error is triggered if you send gain or gamma as a parameter to `getMapUrl()` function because we don't know if requested WMS services support `gain` or `gamma` URL parameters and we don't send them to the services. Because we don't want to deceive the users, we trigger the error instead of requesting the images without these params. Can be seen in:
- `AbstractSentinelHubV1OrV2Layer.ts`
- `AbstractSentinelHubV3Layer.ts`
- `WmsLayer.ts`

In connection to the error, there are changes in  `wms.ts` - the params `gain` and `gamma` are deleted from `OgcGetMapOptions` and not used in `wmsGetMapUrl()`; so we don't send them to the services.

---

Added stories for
- BYOC: `byoc.stories.js` 
 - Envisat Meris: `envisatmeris.eocloud.stories.js`
- legacy: 
  - legacyGetMapFromParams: `legacyGetMapFromParams.stories.js`
  - legacyGetMapFromUrl: `legacyGetMapFromUrl.stories.js`
  - legacyGetMapWmsUrlFromParams: `legacyGetMapWmsUrlFromParams.stories.js`
- S2 L2A: `s2l2a.stories.js`
- Proba-V: `wms.probav.stories.js`

---

Explanation for changing the user-provided parameters in `getMap()` in `AbstractLayer.ts`:

When API type is set to *WMS*, `getMap()` uses `getMapUrl()` with the same provided parameters for getting the url of the image. `getMap()` changes the received image according to provided `gain` and `gamma` *after* it is received.    
An error is thrown in `getMapUrl()` in case `gain` and `gamma` are present, because:
- we don't send `gain` and `gamma` to the services as they may not be supported there and we don't want to deceive the users with returning the image where `gain` and `gamma` were ignored
- if they are supported on the services, `gain` and `gamma` would be applied twice in `getMap()` if they were sent to the services in `getMapUrl()`    

This is a dirty fix, but `gain` and `gamma` need to be removed from the parameters in `getMap()` so the errors in `getMapUrl()` are not triggered.

---

This MR also adds `concurrently ` which allows running 2/more scripts at the same time (in our case `rollup` and `storybook`


src/mapDataManipulation/manipulateGainGamma.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // change the interval of the values from [0, 1] back to [0, 255]
+  newRedValues = newRedValues.map(x => Math.round((255 / 1) * x));
+  newGreenValues = newGreenValues.map(x => Math.round((255 / 1) * x));
+  newBlueValues = newBlueValues.map(x => Math.round((255 / 1) * x));
+
+  const manipulatePixel = function(r: number, g: number, b: number, a: number): object {
+    return { r: newRedValues[r], g: newGreenValues[g], b: newBlueValues[b], a };
+  };
+  return await mapDataManipulation(originalBlob, manipulatePixel);
```
Shooshte, line 54: I would import and reuse `manipulateGain` and `manipulateGamma` functions here. This way you can test (and think) about each manipulation separately, and still have a method that does both where required.

src/mapDataManipulation/manipulateGain.ts
```diff
@@ -0,0 +1,46 @@
+import { mapDataManipulation } from 'src/mapDataManipulation/mapDataManipulation';
+
+// The algorithm works with numbers between 0 and 1, so we must:
+// - change the interval of the values from [0, 255] to [0, 1]
```
Shooshte, line 4: Can we add a check that throws an error if the interval of the values is not between 0 and 255?

src/mapDataManipulation/manipulateGain.ts
```diff
@@ -0,0 +1,46 @@
...
+  offset = offset - factor * minValue;
+
+  let newRedValues = [...Array(256).keys()];
+  let newGreenValues = [...Array(256).keys()];
+  let newBlueValues = [...Array(256).keys()];
+
+  // change the interval of the values from [0, 255] to [0, 1]
+  newRedValues = newRedValues.map(x => (1 / 255) * x);
+  newGreenValues = newGreenValues.map(x => (1 / 255) * x);
+  newBlueValues = newBlueValues.map(x => (1 / 255) * x);
```
Shooshte, line 30: Since this code is used by `manupulateGain`, `manipulateGamma` and `manipulateGainGamma` I think we should write a function that can be reused in all 3 files.

src/mapDataManipulation/manipulateGain.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // change the values according to the algorithm
+  newRedValues = newRedValues.map(x => Math.max(0.0, x * factor + offset));
+  newGreenValues = newGreenValues.map(x => Math.max(0.0, x * factor + offset));
+  newBlueValues = newBlueValues.map(x => Math.max(0.0, x * factor + offset));
+
+  // change the interval of the values from [0, 1] back to [0, 255]
+  newRedValues = newRedValues.map(x => Math.round((255 / 1) * x));
+  newGreenValues = newGreenValues.map(x => Math.round((255 / 1) * x));
+  newBlueValues = newBlueValues.map(x => Math.round((255 / 1) * x));
```
Shooshte, line 40: Since this code is used by `manupulateGain`, `manipulateGamma` and `manipulateGainGamma` I think we should write a function that can be reused in all 3 files.

src/mapDataManipulation/manipulateGain.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const manipulatePixel = function(r: number, g: number, b: number, a: number): object {
+    return { r: newRedValues[r], g: newGreenValues[g], b: newBlueValues[b], a };
+  };
+  return await mapDataManipulation(originalBlob, manipulatePixel);
```
Shooshte, line 45: Since this code is used by `manupulateGain`, `manipulateGamma` and `manipulateGainGamma` I think we should write a function that can be reused in all 3 files.

src/mapDataManipulation/mapDataManipulation.ts
```diff
@@ -0,0 +1,56 @@
+// manipulatePixel is a function that transforms pixel's (R,G,B) values into new (R,G,B) values
+export async function mapDataManipulation(originalBlob: Blob, manipulatePixel: any): Promise<Blob> {
+  let imgObjectUrl: any;
+  try {
+    const imgCanvas = document.createElement('canvas');
```
Shooshte, line 5: I think we should use a unique name for the `imgCanvas` instead of `'canvas'` - since the function is a Promise it might be run in parallel resulting in a race condition that overwrites the previous canvas before `mapDataManipulation` finishes?

zcernigoj: If you are referring to the `canvas` inside of `document.createElement()`, that's the `tagName` of the HTML element, not the `id` / `name` / `class`. We could create a [`HTMLUnknownElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLUnknownElement), but that would be unnecessary work in my opinion. ([MDN docs for `document.createElement()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement))

Shooshte: You are correct. I assumed that this was the reference and not the `tagName`. Thanks for the link

src/mapDataManipulation/manipulateGain.ts
zcernigoj: Talked in-person, decided to simply limit to 0 and 255.

src/mapDataManipulation/manipulateGain.ts
zcernigoj: I refactored / rewrote the things quite a bit, but used that advice while at it.

src/mapDataManipulation/manipulateGain.ts
zcernigoj: I refactored / rewrote the things quite a bit, but used that advice while at it.

src/mapDataManipulation/manipulateGain.ts
zcernigoj: I refactored / rewrote the things quite a bit, but used that advice while at it.

zcernigoj:
I refactored / rewrote / reorganized the things quite a bit. It should be now easier to add predefined effect function without duplicating the code that prepares things.

src/mapDataManipulation/manipulateGainGamma.ts
Shooshte: Resolving, will check the refactored code

src/mapDataManipulation/manipulateGain.ts
zcernigoj: resolving so i can merge

src/mapDataManipulation/manipulateGain.ts
zcernigoj: resolving so i can merge

src/mapDataManipulation/manipulateGain.ts
zcernigoj: resolving so i can merge

PR #83, opened 2020-05-08
Fix/data fusion evalscripturl

sinergise-anze:
This MR adds support for `evalscriptUrl` to `ProcessingDataFusionLayer`. 

I have rearranged the code a bit. `convertEvalscriptToV3IfNeeded` is also changed a bit, so that conversion is marked as "done" if the evalscript starts with `//VERSION=3` already.

Additionally, `mosaickingOrder` is not fetched from the service when `instanceId` or `layerId` are not available (fixes storybook `GetMapProcessingEvalscripturlVersion2`).

### DIFF
--- a/stories/datafusion.stories.js
+++ b/stories/datafusion.stories.js
@@ -62,13 +62,13 @@ export const getMapProcessing = () => {
     const layers = [
       {
         layer: layerS2L2A,
-        id: 'l2a',
+        id: 's2l2a',
         fromTime: new Date(Date.UTC(2020, 1 - 1, 1, 0, 0, 0)),
         toTime: new Date(Date.UTC(2020, 2 - 1, 26, 0, 0, 0)),
       },
       {
         layer: layerS2L1C,
-        id: 'l1c',
+        id: 's2l1c',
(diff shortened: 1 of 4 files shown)


stories/datafusion.stories.js
```diff
@@ -79,16 +79,21 @@ export const getMapProcessing = () => {
         //VERSION=3
         var setup = () => ({
           input: [
-            {datasource: "l2a", bands:["B02", "B03", "B04"], units: "REFLECTANCE", mosaicking: "ORBIT"},
-            {datasource: "l1c", bands:["B02", "B03", "B04"], units: "REFLECTANCE"}],
+            {datasource: "s2l2a", bands:["B02", "B03", "B04"], units: "REFLECTANCE", mosaicking: "ORBIT"},
```
dgostencnik, line 82: Extreme nitpicking: title of  story should be changed to `Data fusion - Sentinel-2 L1C and Sentinel-2 L2A` as Landsat 8 is not used in this example. fftr

sinergise-anze: Thank you, fixed!

PR #84, opened 2020-05-09
Feature/tiles extra info

Ardweaden:
This MR enriches tiles from  `findTiles` with `links` and additional metadata. Specifically, metadata and links which are displayed in EOBrowser.

I kept the `links`  fields optional.
https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields

### DIFF
--- a/src/layer/AbstractSentinelHubV3WithCCLayer.ts
+++ b/src/layer/AbstractSentinelHubV3WithCCLayer.ts
@@ -76,9 +76,11 @@ export class AbstractSentinelHubV3WithCCLayer extends AbstractSentinelHubV3Layer
       maxcc: this.maxCloudCoverPercent,
     };
   }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   protected extractFindTilesMeta(tile: any): Record<string, any> {
     return {};
   }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   protected getTileLinks(tile: Record<string, any>): Link[] {
     return [];
   }
--- a/src/layer/AbstractSentinelHubV3WithCCLayer.ts
+++ b/src/layer/AbstractSentinelHubV3WithCCLayer.ts
@@ -59,14 +59,7 @@ export class AbstractSentinelHubV3WithCCLayer extends AbstractSentinelHubV3Layer
       tiles: response.data.tiles.map(tile => ({
         geometry: tile.dataGeometry,
         sensingTime: moment.utc(tile.sensingTime).toDate(),
-        meta: {
-          cloudCoverPercent: tile.cloudCoverPercentage,
-          tileId: tile.id,
-          MGRSLocation: tile.dataUri
-            .split('/')
-            .slice(4, 7)
-            .join(''),
-        },
+        meta: this.extractFindTilesMeta(tile),
         links: this.getTileLinks(tile),
       })),
       hasMore: response.data.hasMore,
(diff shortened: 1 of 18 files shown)


src/layer/EnvisatMerisEOCloudLayer.ts
```diff
@@ -23,4 +23,14 @@ export class EnvisatMerisEOCloudLayer extends AbstractSentinelHubV1OrV2Layer {
       description,
     });
   }
+
+  protected getTileLinks(tile: Record<string, any>): Link[] {
+    return [
+      {
+        href: tile.pathFragment,
+        rel: 'self',
+        title: 'EOCloudPath',
```
sinergise-anze, line 32: I'm a bit torn here... On the one hand, it would be nice to use part of STAC definition here, but on the other side it is not really fitting... `rel` will always be `self`, and we can't use `title` as ID. We will need to know about the different link types though.

Maybe we can define `Link` like this?
```
{
  href: tile.pathFragment,
  type: 'eocloud',
}
```

Additionally, I assume that `href` always contains an absolute URL? If not, I think a better name would be `target`, `path`, `link` or similar...

src/layer/Landsat5EOCloudLayer.ts
```diff
@@ -24,4 +25,26 @@ export class Landsat5EOCloudLayer extends AbstractSentinelHubV1OrV2WithCCLayer {
...
+        title: 'EOCloudPath',
+      },
+      {
+        href: `${tile.previewUrl.replace('eocloud', 'creodias')}.JPG`,
+        rel: 'self',
+        title: 'Preview',
+      },
+    ];
+  }
+
```
sinergise-anze, line 43: IIUC, this is also not needed, as it is defined on the parent class already? 

(the same below)

Ardweaden: I'm not sure what you mean with this.

src/layer/AbstractSentinelHubV3WithCCLayer.ts
```diff
@@ -77,4 +76,12 @@ export class AbstractSentinelHubV3WithCCLayer extends AbstractSentinelHubV3Layer
...
     };
   }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected extractFindTilesMeta(tile: any): Record<string, any> {
+    return {};
+  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected getTileLinks(tile: Record<string, any>): Link[] {
+    return [];
+  }
```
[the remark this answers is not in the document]
Ardweaden: Should be fixed.

src/layer/AbstractSentinelHubV3WithCCLayer.ts
```diff
@@ -77,4 +76,12 @@ export class AbstractSentinelHubV3WithCCLayer extends AbstractSentinelHubV3Layer
       maxcc: this.maxCloudCoverPercent,
     };
   }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected extractFindTilesMeta(tile: any): Record<string, any> {
+    return {};
+  }
```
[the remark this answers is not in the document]
Ardweaden: Sorry about that.

src/layer/EnvisatMerisEOCloudLayer.ts
Ardweaden: It was defined as:
```
{
  target: string,
  type: LinkType,
}
```

src/layer/Landsat5EOCloudLayer.ts
sinergise-anze: Nevermind... :)

PR #85, opened 2020-05-11
Feature/upsampling defaults

dgostencnik:
This is pretty much the same problem as  with `mosaickingOrder` and `maxCloudCoverage`. Default values for `upsampling` and `downsampling` can be defined in dashboard and that can lead to different images when using wms and processing.  Nice example is eob3 where in dashboard  upsampling for all S2L1C layers is set to `BICUBIC` , but `NEAREST` is used when calling getMap (you can see this as blurry vs. pixelated image when using small zoom). To "solve" this, upsampling and downsampling were moved to layer properties.

### DIFF
--- a/src/layer/AbstractSentinelHubV1OrV2Layer.ts
+++ b/src/layer/AbstractSentinelHubV1OrV2Layer.ts
@@ -81,7 +81,7 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
       additionalParameters.upsampling = this.upsampling;
     }
     if (this.downsampling) {
-      additionalParameters.priority = this.downsampling;
+      additionalParameters.downsampling = this.downsampling;
     }
     return additionalParameters;
   }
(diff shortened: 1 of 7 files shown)


src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -64,12 +73,17 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
-        priority: this.mosaickingOrder,
-      };
+      additionalParameters.priority = this.mosaickingOrder;
     }
-    return {};
+    if (this.upsampling) {
+      additionalParameters.upsampling = this.upsampling;
+    }
+    if (this.downsampling) {
+      additionalParameters.priority = this.downsampling;
```
zcernigoj, line 84: Typo: Should be `downsampling` instead of `priority `

dgostencnik: Was changed.

zcernigoj: Resolving.

PR #88, opened 2020-05-13
Add Sentinel-2 L1C's previews to L2A

Ardweaden:
Adds previews for Sentinel-2 L2A (same as L1C)

https://trello.com/c/g719ReMy/431-s2l2a-doesnt-show-preview-image

### DIFF
--- a/src/layer/S2L2ALayer.ts
+++ b/src/layer/S2L2ALayer.ts
@@ -17,6 +17,12 @@ export class S2L2ALayer extends AbstractSentinelHubV3WithCCLayer {
         target: tile.dataUri,
         type: LinkType.AWS,
       },
+      {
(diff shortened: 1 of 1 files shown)


src/layer/S2L2ALayer.ts
```diff
@@ -17,6 +17,12 @@ export class S2L2ALayer extends AbstractSentinelHubV3WithCCLayer {
         target: tile.dataUri,
         type: LinkType.AWS,
       },
+      {
```
sinergise-anze, line 20: It would be nice if there was a comment here:
```
// S-2 L2A doesn't have previews, but we can use corresponding L1C ones instead:
```

Fftr.

Ardweaden: I have added it.

PR #91, opened 2020-05-13
Feature/update BYOC layer on makeLayers

dgostencnik:
Some of layer properties are lazily fetched from service on getMap and few other functions. For BYOC layers such properties are collectionID and location. However fetching those properties on getMap is not enough. CollectionId is needed when constructing search form and to create search functions (unless wrong assumption is made that there can be only one collection in each instance). 
IMO collectionId should be provided by service in getCapabilities, but since this discussion was ended some time ago by higher authority, this is an attempt to set collectionId when making layers from capabilities.

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -138,25 +138,34 @@ export class LayersFactory {
     const filteredLayersInfos =
       filterLayers === null ? layersInfos : layersInfos.filter(l => filterLayers(l.layerId, l.dataset));
 
-    return filteredLayersInfos.map(({ layerId, dataset, title, description }) => {
-      if (!dataset) {
-        return new WmsLayer({ baseUrl, layerId, title, description });
-      }
+    return Promise.all(
+      filteredLayersInfos.map(async ({ layerId, dataset, title, description }) => {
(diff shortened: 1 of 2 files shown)


src/layer/LayersFactory.ts
```diff
@@ -138,25 +138,34 @@ export class LayersFactory {
...
+          evalscript: null,
+          evalscriptUrl: null,
+          dataProduct: null,
+          title,
+          description,
+        });
+
+        if (layer instanceof BYOCLayer) {
+          await layer.updateLayerFromServiceIfNeeded();
+        }
```
sinergise-anze, line 164: I understand the reasoning behind updating the layer here, but the convention elsewhere is that we lazy-load the needed data only when needed. In other words, every method in `BYOCLayer` that needs a `collectionId` should perform `updateLayerFromServiceIfNeeded` (actually, I thought it already did?).

Also, while I didn't test it, I think constructing a BYOCLayer with only `instanceId` and `layerId` would not work then?
```
  const layer = new BYOCLayer({ instanceId: ..., layerId: ... });
```

I would suggest using lazy-loading when needed, instead of on construction...

PR #92, opened 2020-05-13
Fix/convert byoc evalscript

dgostencnik:
When converting evalscripts used by BYOC layers, additional parameter `byocCollectionId` is required by service.

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -397,7 +397,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     return res.data;
   }
 
-  protected getConvertScriptBaseUrl() {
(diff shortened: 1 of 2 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -397,4 +396,9 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const res = await axios.post(url, evalscript, requestConfig);
     return res.data;
   }
+
+  protected getConvertScriptBaseUrl(): string {
```
sinergise-anze, line 400: Just for consistency, can we maybe name this `getConvertEvalscriptBaseUrl`? Fftr.

dgostencnik: Was renamed

PR #94, opened 2020-05-19
Fix/add gain gamma readme

zcernigoj:
This MR adds documentation for gain and gamma to `readme.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,9 +177,9 @@ Gain and gamma effects are supported only when the blob is retrieved, because th
 ```
 
 As mentioned above, both of the images above should be _exactly_ the same, no matter which API type is used. 
-This is also true when using gain and gamma effects.
+This is still true when applying gain and gamma effects.
 
-When retrieving an image URL with gain and gamma applied, an error is thrown.
+When retrieving an image URL with gain and gamma applied, an error is thrown, because the retrieved URL points directly to the image on the services with no applied effects.
 
 ## Searching for data
 
--- a/README.md
+++ b/README.md
@@ -159,6 +159,28 @@ It is also possible to determine whether a layer supports a specific ApiType:
   };
 ```
 
+Gain and gamma effects are supported only when the blob is retrieved, because the effects are applied in the library. 
+
+```javascript
+  const getMapParamsWithGainAndGamma = {
+    bbox: bbox,
+    fromTime: new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0)),
+    toTime: new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59)),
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -159,6 +159,28 @@ It is also possible to determine whether a layer supports a specific ApiType:
   };
 ```
 
+Gain and gamma effects are supported only when the blob is retrieved, because the effects are applied in the library.
```
sinergise-anze, line 162: The way the effects are applied is a good approach and it was quite a lot of work, but - I don't think users care about that... :) It is (and should be) transparent to them. For this reason, I would suggest simply stating:
```
Effects `gain` and `gamma` can also be applied:
```

README.md
```diff
@@ -159,6 +159,28 @@ It is also possible to determine whether a layer supports a specific ApiType:
...
+    gain: 1.2,
+    gamma: 0.9
+  };
+  const imageBlob = await layer.getMap(getMapParamsWithGainAndGamma, ApiType.WMS);
+  const imageBlob2 = await layer.getMap(getMapParamsWithGainAndGamma, ApiType.PROCESSING);
+```
+
+As mentioned above, both of the images above should be _exactly_ the same, no matter which API type is used. 
+This is still true when applying gain and gamma effects.
+
```
sinergise-anze, line 181: I would suggest removing this paragraph as its information is already conveyed elsewhere.

README.md
```diff
@@ -159,6 +159,28 @@ It is also possible to determine whether a layer supports a specific ApiType:
...
+When retrieving an image URL with gain and gamma applied, an error is thrown, because the retrieved URL points directly to the image on the services with no applied effects.
```
sinergise-anze, line 182: This is good, I would just suggest adding functino name to make it clearer what we are talking about:
```
When retrieving an image URL (`getMapUrl`) with gain and gamma applied, an error is thrown, because the retrieved URL points directly to the image on the services with no applied effects.
```

README.md
zcernigoj: I didn't know how detailed the description about gain and gamma being applied by the library should be. 
Thanks, this looks better.

README.md
zcernigoj: Thanks, removed.

README.md
zcernigoj: Thanks, added the function name.

README.md
zcernigoj: Resolving so I can merge

README.md
zcernigoj: Resolving so I can merge

README.md
zcernigoj: Resolving so I can merge

PR #95, opened 2020-05-20
Fix/ pass maxcc from legacy functions to makeLayer(s)

zcernigoj:
When using legacy functions, `maxcc` parameter was ignored and thus in Playground and EOBrowser the layers were created using the `maxCloudCoverPercent` from the layers definitions from services (mostly 100 on Sentinel Hub and 20 on EOCloud).
This MR passes the `maxcc` parameter (is implemented that it could also pass others, like the ones for Sentinel-1) to `makeLayer` and `makeLayers` and is used to override the default values from layers definitions.

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -97,9 +97,15 @@ export class LayersFactory {
   public static async makeLayer(
     baseUrl: string,
     layerId: string,
+    overrideSHLayerParams: Record<string, any> | null,
     reqConfig?: RequestConfiguration,
   ): Promise<AbstractLayer> {
-    const layers = await LayersFactory.makeLayers(baseUrl, (lId: string) => lId === layerId, reqConfig);
+    const layers = await LayersFactory.makeLayers(
+      baseUrl,
+      (lId: string) => lId === layerId,
+      overrideSHLayerParams,
+      reqConfig,
+    );
     if (layers.length === 0) {
       return null;
     }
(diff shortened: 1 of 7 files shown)


src/legacyCompat.ts
```diff
@@ -35,9 +35,16 @@ export async function legacyGetMapFromParams(
...
+  const {
+    layers,
+    evalscript,
+    evalscriptUrl,
+    evalsource,
+    getMapParams,
+    otherLayerParams,
+  } = parseLegacyWmsGetMapParams(wmsParams);
+
+  console.log('lgmfromparams', { getMapParams, otherLayerParams });
```
dgostencnik, line 47: this should be removed

src/layer/LayersFactory.ts
```diff
@@ -109,17 +115,18 @@ export class LayersFactory {
   public static async makeLayers(
     baseUrl: string,
     filterLayers: Function | null = null,
+    overrideSHLayerParams?: Record<string, any>,
```
dgostencnik, line 118: I don't think this is the most accurate name for this variable as you're setting default constructor parameters here and not overriding some values from service. Perhaps `defaultConstructorParameters` or something would be better? Although this is just semantic nitcpicking as intend is clear enaugh. fftr.

zcernigoj: Me and @sinergise-anze came up with this name after discussing the problem, so I would leave it as is.

src/legacyCompat.ts
zcernigoj: Thanks, removed.

src/layer/LayersFactory.ts
dgostencnik: :+1:

sinergise-anze: Actually, we came up with `overrideLayerParams` which is something else entirely... :smile: 

To me `override` seems fitting, because we are overriding the values we get from the layer definition on the service. `overrideConstructorParams` maybe?

zcernigoj: True, my bad, original was `overrideLayerParams`, I just added `SH` since currently only legacy functions use that and only for SH layers.

To me `overrideSomething` seems ok, since the layer classes default behaviour is to use the values from the services (and it seems that this wasn't obvious before). But I don't care much what comes after that. Will change to `overrideConstructorParams`.

zcernigoj: Was changed

PR #96, opened 2020-05-20
Temporary fix to use WMS when there's a dataProduct on the layer

ignasieb:
Temporal fix to only use WMS when a layer has a dataProduct on it. This should be reverted when the core team fixes or converts all dataProducts to V3

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -225,9 +225,8 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   public supportsApiType(api: ApiType): boolean {
     if (this.dataProduct) {
       return api === ApiType.WMS;
-    } else {
-      return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
     }
+    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
   }
 
   protected getWmsGetMapUrlAdditionalParameters(): Record<string, any> {
@@ -494,6 +493,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   public async updateLayerFromServiceIfNeeded(reqConfig?: RequestConfiguration): Promise<void> {
     const layerParams = await this.fetchLayerParamsFromSHServiceV3(reqConfig);
     this.legend = layerParams['legend'] ? layerParams['legend'] : null;
+    // this is a hotfix for `supportsApiType()` not having enough information - should be fixed properly later:
     this.dataProduct = layerParams['dataProduct'] ? layerParams['dataProduct'] : null;
   }
   protected getConvertEvalscriptBaseUrl(): string {
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -491,9 +491,10 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     return res.data;
   }
 
-  public async updateLayerFromServiceIfNeeded(reqConfig: RequestConfiguration): Promise<void> {
+  public async updateLayerFromServiceIfNeeded(reqConfig?: RequestConfiguration): Promise<void> {
     const layerParams = await this.fetchLayerParamsFromSHServiceV3(reqConfig);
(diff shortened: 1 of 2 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -487,9 +491,10 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     return res.data;
   }
 
-  public async updateLayerFromServiceIfNeeded(reqConfig: RequestConfiguration): Promise<void> {
+  public async updateLayerFromServiceIfNeeded(reqConfig?: RequestConfiguration): Promise<void> {
```
sinergise-anze, line 494: This is probably a typo? `fetchLayerParamsFromSHServiceV3` function signature also requests `reqConfig` to be defined.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -223,7 +223,11 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   }
 
   public supportsApiType(api: ApiType): boolean {
-    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
+    if (this.dataProduct) {
+      return api === ApiType.WMS;
+    } else {
+      return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
+    }
```
sinergise-anze, line 230: Just a minor nitpick... It is usually easier to follow code from top to bottom, than it is to understand the nesting levels. Because of that I would suggest removing unneeded `else` block:
```
    if (this.dataProduct) {
      return api === ApiType.WMS;
    };
    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
```

Fftr.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -487,9 +491,10 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     return res.data;
   }
 
-  public async updateLayerFromServiceIfNeeded(reqConfig: RequestConfiguration): Promise<void> {
+  public async updateLayerFromServiceIfNeeded(reqConfig?: RequestConfiguration): Promise<void> {
     const layerParams = await this.fetchLayerParamsFromSHServiceV3(reqConfig);
     this.legend = layerParams['legend'] ? layerParams['legend'] : null;
+    this.dataProduct = layerParams['dataProduct'] ? layerParams['dataProduct'] : null;
```
sinergise-anze, line 497: This doesn't look like a correct place to update `dataProduct` to me... If nothing else, there should be no difference between `evalscript` and `dataProduct`, so if one is set (from the service), the other one should be too. I have checked where this is used but failed to see where this is needed, as it is updated in `getMap` anyway - is there a special case that I missed?

ignasieb: Yes, in getMap gets updated but the check of the apiType happens before getMap is called in EOB3. So the idea is to updateLayer and then check supportApiType to decide whether WMS or Processing is used.

src/layer/AbstractSentinelHubV3Layer.ts
ignasieb: Well because it's a public method I thought it would be better to have the request configuration as optional. For all protected methods this is mandatory but since this one is public I thought it may be a better option.

sinergise-anze: I see... In that case all of them should be optional, until we make a check if the `reqConfig` was supplied or not. But this is probably not within the scope of this MR, so we can leave it as it is... resolving.

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: Still, this doesn't seem like the proper way to do it... Let's discuss live how to best solve this on Monday.

sinergise-anze: As discussed, we can reolve this, but I would suggest adding a comment:
```
this.legend = ...
// this is a hotfix for `supportsApiType()` not having enough information - should be fixed properly later:
this.dataProduct = ...
```

Fftr.

ignasieb: Done. Thanks Anze

PR #97, opened 2020-05-21
fix/ makeLayersSHv12: only use overrideConstructorParams if its not null + mapDataManipulation return correct blob type

zcernigoj:
2 small fixes:

- makeLayersSHv12 only checked if maxCloudCoverPercent from  overrideConstructorParams is not null but not if overrideConstructorParams is not null

- mapDataManipulation returned `image/jpeg` instead of the type that the provided blob had.

### DIFF
--- a/stories/s2l2a.stories.js
+++ b/stories/s2l2a.stories.js
@@ -641,6 +641,7 @@ export const getMapProcessingGainGammaCheckTransparency = () => {
   const wrapperEl = document.createElement('div');
   wrapperEl.innerHTML = '<h2>S2L2A getMapProcessingGainGamma; Check transparency</h2>';
   wrapperEl.innerHTML += '<h4>no gain/gamma | gain | gamma | gain and gamma</h4>';
+  wrapperEl.innerHTML += '<p>Note: Whole images can be transparent if there is no data to show</p>';
   wrapperEl.style.backgroundColor = 'lightgreen';
   wrapperEl.insertAdjacentElement('beforeend', imgNoGainGamma);
   wrapperEl.insertAdjacentElement('beforeend', imgGainIs2);
@@ -676,8 +677,8 @@ export const getMapProcessingGainGammaCheckTransparency = () => {
 
     const getMapParams = {
       bbox: bbox3857,
-      fromTime: new Date(Date.UTC(2020, 5 - 1, 20, 0, 0, 0)),
-      toTime: new Date(Date.UTC(2020, 5 - 1, 20, 23, 59, 59)),
+      fromTime: new Date(Date.UTC(2020, 4 - 1, 15, 0, 0, 0)),
+      toTime: new Date(Date.UTC(2020, 4 - 1, 15, 23, 59, 59)),
       width: 512,
       height: 512,
       format: MimeTypes.PNG,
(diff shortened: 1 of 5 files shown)


stories/s2l2a.stories.js
```diff
@@ -614,6 +621,93 @@ export const getMapProcessingGainGamma = () => {
   return wrapperEl;
 };
 
+export const getMapProcessingGainGammaCheckTransparency = () => {
```
dgostencnik, line 624: What is expected result for this? 4 fully transparent images?

zcernigoj: Sort of. The 100% correct would be the same as this curl request (only part of the image has data). But 4 fully transparent images are also correct - they indicate that datamask works (sets transparency if there is no data, and apparently there is no data for the request that the `getMap` makes here)

```bash
curl 'https://services.sentinel-hub.com/api/v1/process' \
  -H 'authority: services.sentinel-hub.com' \
  -H 'pragma: no-cache' \
  -H 'cache-control: no-cache' \
  -H 'accept: */*' \
  -H 'authorization: Bearer ...' \
  -H 'user-agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.138 Safari/537.36' \
  -H 'content-type: application/json' \
  -H 'origin: http://localhost:3000' \
  -H 'sec-fetch-site: cross-site' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-dest: empty' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8,sl;q=0.7' \
  --data-binary '{"input":{"bounds":{"bbox":[1174072.7544603075,5009377.085697314,1252344.2714243277,5087648.602661333],"properties":{"crs":"http://www.opengis.net/def/crs/EPSG/0/3857"}},"data":[{"dataFilter":{"timeRange":{"from":"2020-05-20T00:00:00.000Z","to":"2020-05-20T23:59:59.999Z"},"mosaickingOrder":"mostRecent","previewMode":"EXTENDED_PREVIEW","maxCloudCoverage":100},"processing":{"upsampling":"BICUBIC"},"type":"S2L1C"}]},"output":{"width":512,"height":512,"responses":[{"identifier":"default","format":{"type":"image/png"}}]},"evalscript":"//VERSION=3\nlet minVal = 0.0;\nlet maxVal = 0.4;\n\nlet viz = new HighlightCompressVisualizer(minVal, maxVal);\n\nfunction setup() {\n   return {\n    input: [\"B04\", \"B03\", \"B02\",\"dataMask\"],\n    output: { bands: 4 }\n  };\n}\n\nfunction evaluatePixel(samples) {\n    let val = [samples.B04, samples.B03, samples.B02,samples.dataMask];\n    return viz.processList(val);\n}"}' \
  --compressed
```

dgostencnik: I would perhaps make more sense to use some other location where only part of image would be transparent. Or just add in story title that results are fully transparent images.

zcernigoj: Thanks, I changed the date to one that there is some data on the image; and added a note that images may be fully transparent if there is no data.

PR #98, opened 2020-05-21
Feature: getAvailableBands method for BYOC layer

Ardweaden:


src/layer/BYOCLayer.ts
```diff
@@ -183,4 +183,16 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
...
+    const url = `https://services.sentinel-hub.com/api/v1/metadata/collection/CUSTOM/${this.collectionId}`;
+    const headers = { Authorization: `Bearer ${getAuthToken()}` };
+    const res = await axios.get(url, {
+      responseType: 'json',
+      headers: headers,
+      useCache: true,
+      ...getAxiosReqParams(reqConfig),
+    });
+    return res.data.bands;
+  }
```
zcernigoj, line 197: Nitpicking, this works well too, but it doesn't save something that is a property of a layer to the layer.

I would make `bands` a property of `BYOCLayer` (would need to add it to `ConstructorParameters` and `constructor()`). Then I would save it the same way as `locationId` in `updateLayerFromServiceIfNeeded()` (with the checks if it's not null etc.). 

In this function (`getAvailableBands`) I would call `updateLayerFromServiceIfNeeded()` and then return `this.bands`.
Similar to what `getFindDates...()` methods are doing.

zcernigoj: Or, other solution would be to just add saving the `bands` in `updateLayerFromServiceIfNeeded()` and then in the project where you need them, do: 

```javascript
let byocLayer = new BYOCLayer(...);
byocLayer.updateLayerFromServiceIfNeeded();
const bands = byocLayer.bands;
```

Ardweaden: I answered on the EOB3 MR: https://git.sinergise.com/team-6/eobrowser3/merge_requests/166#note_115691

zcernigoj: Apologies, true, I mixed the dataset and layer. Bands are not really a property of a layer, since you can create numerous different layers from them. Resolving.

src/layer/BYOCLayer.ts
```diff
@@ -183,4 +183,16 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
   protected getConvertEvalscriptBaseUrl(): string {
     return `${super.getConvertEvalscriptBaseUrl()}&byocCollectionId=${this.collectionId}`;
   }
+
+  public async getAvailableBands(reqConfig?: RequestConfiguration): Promise<void> {
+    const url = `https://services.sentinel-hub.com/api/v1/metadata/collection/CUSTOM/${this.collectionId}`;
```
zcernigoj, line 188: I would still check if `this.collectionId` is set and call `updateLayerFromServiceIfNeeded()` if it's not. Just to be on a safe side.

Ardweaden: True, it was added.

zcernigoj: :+1: Resolving.

PR #99, opened 2020-05-22
Only apply effects if params gain or gamma are set

sinergise-anze:
The original issue was that `console.error` message was being output when running tests. The reason was that node.js doesn't support Canvas natively, so the image manipulation methods failed. Still, they shouldn't have been run in the first place, because parameters `gain` and `gamma` were not set in tests. 

This MR fixes the simplest of the problems above - image manipulation should not happen if parameters `gain` and `gamma` are not set.

Additionally, we avoid changing the input structure (`params`) in `AbstractLayer::getMap()`.


sinergise-anze:
> `if (params.gain || params.gamma) {`
> 
> This won't work if both are 0 (defined, not null, and are 0)

True, thank you for noticing! Fixed.

PR #100, opened 2020-05-22
Add test for parseLegacyWmsGetMapParams

sinergise-anze:
Also fixes: 
- `GetMapParams` should be exported
- limit preview to max `2` (`3` is allowed for compatibility reason, but we should change it to `2`)
- fix data type for transparent (always boolean internally)

Note: `unknown` might not have correct content (it should probably be empty), but this is a bigger issue and I didn't want to dive into it ATM, so I have just left a comment to warn future readers.

### DIFF
--- a/src/layer/__tests__/legacy.ts
+++ b/src/layer/__tests__/legacy.ts
@@ -24,6 +24,8 @@ test('parseLegacyWmsGetMapParams with evalscripturl', () => {
     transparent: '1',
     bgcolor: '00000000',
     preview: 3,
+    gain: 0.7,
+    gamma: 0.9,
   };
   const {
     layers,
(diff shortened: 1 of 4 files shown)


src/layer/__tests__/legacy.ts
```diff
@@ -0,0 +1,67 @@
+import 'jest-setup';
+import moment from 'moment';
+
+import { parseLegacyWmsGetMapParams, BBox, CRS_EPSG3857, GetMapParams, PreviewMode } from 'src';
+
+test('parseLegacyWmsGetMapParams with evalscripturl', () => {
+  const evalscriptUrlOriginal =
+    'https://gist.githubusercontent.com/sinergise-anze/33fe78d9b1fd24d656882d7916a83d4d/raw/295b9d9f033c7e3f1e533363322d84846808564c/data-fusion-evalscript.js';
+  const wmsParams = {
```
zcernigoj, line 9: Minor: I would also add `gain` and `gamma` parameters as well.

sinergise-anze: Added.

zcernigoj: Nice, resolving.

PR #104, opened 2020-06-02
Feature/request config timeout

Shooshte:
Proposition on how to handle class method timeouts. Creating apull request to review before adding the decorator to all exposed async class methods

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,25 @@ We can always use layer to search for data availability:
   const datesS1 = await layerS1.findDatesUTC(bbox, fromTime, toTime);
 ```
 
+## Request timeout
+
+You can specify a timeout in milliseconds for network requests. This will cancel all the network requests triggered by the method after the specified time frame. When not specified the method will not time out.
+
+```typescript
+const requestConfig = {
+  timeout: 5000,
+};
+
+try {
+  const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestConfig);
+  const dates = await layer.findDatesUTC(bbox, fromTime, toTime, requestConfig);
+  const stats = await layer.getStats(getStatsParams, requestConfig);
+  const tiles = await layer.findTiles(bbox, fromTime, toTime, null, null, requestConfig);
+} catch (err) {
+  throw err;
+}
+```
+
 
 ## Cancelling requests
 
--- a/README.md
+++ b/README.md
@@ -190,7 +190,17 @@ We can always use layer to search for data availability:
 
 ## Requests configuration
 
+You can specify that network requests should be retried by passing the number of retries for network requests used by the method. The default of `retries` is `null` (disabled).
+
+```typescript
+const requestsConfig = {
+  retries: 4
+}
+```
+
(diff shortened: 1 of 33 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -30,7 +30,12 @@ export class AbstractLayer {
     this.legendUrl = legendUrl;
   }
 
-  public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
+  @timeoutWrapper(2)
+  public async getMap(
+    params: GetMapParams,
+    api: ApiType,
+    requestsConfig?: RequestConfiguration,
```
sinergise-anze, line 37: It is difficult to review this MR because so many (unrelated) things have changed. Maybe a good idea would be to implement just the timeout functionality, and leave cleanup tasks for another MR, so they can be properly checked too?

src/layer/AbstractLayer.ts
```diff
@@ -30,7 +30,12 @@ export class AbstractLayer {
     this.legendUrl = legendUrl;
   }
 
-  public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
+  @timeoutWrapper(2)
```
sinergise-anze, line 33: As we have discussed recently, I think it would be better to avoid using the decorators altogether, not make it more explicit how the `reqConfig` parameter changes. 

The suggestion was:
```typescript
public async getMap(params: GetMapParams, api: ApiType, reqConfigWithTimeout?: RequestConfiguration): Promise<Blob> {
  return await ensureTimeout(reqConfigWithTimeout, async (reqConfig) => {
    ...
  })
}
```

Then `ensureTimeout` can look similar to this:
```javascript
async function ensureTimeout(originalReqConfig, innerFunc) {
  if (!originalReqConfig) {
    return await innerFunc(originalReqConfig);
  }

  // setTimeout, clearTimeout,...

  return newReqConfig;
}
```

This would make implementation much easier (imho) and explicit about what is going on. What do you think?

README.md
```diff
@@ -191,100 +204,121 @@ When retrieving an image URL (via `getMapUrl()`) with gain and gamma applied, an
...
+  const tiles = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+} catch (err) {
+  throw err;
+}
+```
 
-To do so a token needs to be created and passed through a requests configuration object. Other config such as retries can also be defined there.
+You can also cancel requests when searching/fetching data. To do so a token needs to be created and passed through the requests configuration object.
 
 In the example below, a cancel token is passed inside the configuration request object. The timeout will cancel the requests after 500 miliseconds, throwing an exception.
```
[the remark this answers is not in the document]
Shooshte: Resolving this, opened a [different MR](https://github.com/sentinel-hub/sentinelhub-js/pull/105/files)

Shooshte:
Closing this, opened a new [WIP PR](https://github.com/sentinel-hub/sentinelhub-js/pull/105/files)

PR #108, opened 2020-06-11
Feature/basic color manipulation

zcernigoj:
Added support for changing `red`, `green`, `blue` ranges (on what what ranges can values for those colors be)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,8 @@ It is also possible to determine whether a layer supports a specific ApiType:
 When requesting an image, effects can be applied to visually improve the image.
 To apply the effects, the `effects` param in `getMapParams` should be present, containing the desired effects.
 Supported effects are `gain`, `gamma`, `redRange`, `greenRange` and `blueRange`. 
-All of them accept the values between 0 and 1.
+Effects `gain` and `gamma` accept values greater than 0. 
+Effects `redRange`, `greenRange` and `blueRange` accept the values between 0 and 1.
 Setting values to `redRange`, `greenRange` and `blueRange` limits the values that pixels can have for red, green and blue color component respectively.
 
 Effects are applied by the library (client-side) and are thus only available when the blob is retrieved (`getMap`) and not through the URL (`getMapUrl`).
--- a/README.md
+++ b/README.md
@@ -172,8 +172,8 @@ It is also possible to determine whether a layer supports a specific ApiType:
 When requesting an image, effects can be applied to visually improve the image.
 To apply the effects, the `effects` param in `getMapParams` should be present, containing the desired effects.
 Supported effects are `gain`, `gamma`, `redRange`, `greenRange` and `blueRange`. 
-Effects `gain` and `gamma` accept values greater than 0. 
-Effects `redRange`, `greenRange` and `blueRange` accept the values between 0 and 1.
+Effects `gain` and `gamma` accept values equal or greater than 0. 
+Effects `redRange`, `greenRange` and `blueRange` accept the values between 0 and 1, including both 0 and 1.
 Setting values to `redRange`, `greenRange` and `blueRange` limits the values that pixels can have for red, green and blue color component respectively.
 
 Effects are applied by the library (client-side) and are thus only available when the blob is retrieved (`getMap`) and not through the URL (`getMapUrl`).
--- a/README.md
+++ b/README.md
@@ -171,7 +171,9 @@ It is also possible to determine whether a layer supports a specific ApiType:
 
 When requesting an image, effects can be applied to visually improve the image.
 To apply the effects, the `effects` param in `getMapParams` should be present, containing the desired effects.
-Supported effects are gain and gamma. 
+Supported effects are `gain`, `gamma`, `redRange`, `greenRange` and `blueRange`. 
+All of them accept the values between 0 and 1.
+Setting values to `redRange`, `greenRange` and `blueRange` limits the values that pixels can have for red, green and blue color component respectively.
 
 Effects are applied by the library (client-side) and are thus only available when the blob is retrieved (`getMap`) and not through the URL (`getMapUrl`).
 
@@ -188,7 +190,10 @@ When retrieving an image URL (via `getMapUrl()`) with effects applied, an error 
     format: MimeTypes.JPEG,
     effects: {
       gain: 1.2,
-      gamma: 0.9
+      gamma: 0.9,
+      redRange: {from: 0.2, to: 0.8},
+      greenRange: {from: 0.2, to: 0.8},
+      blueRange: {from: 0.2, to: 0.8},
     }
   };
 
--- a/src/mapDataManipulation/const.ts
+++ b/src/mapDataManipulation/const.ts
@@ -1,9 +1,14 @@
+export type ColorRange = {
+  from: number;
+  to: number;
+};
+
 export type Effects = {
   gain?: number;
   gamma?: number;
-  redRange?: number[];
-  greenRange?: number[];
-  blueRange?: number[];
+  redRange?: ColorRange;
+  greenRange?: ColorRange;
+  blueRange?: ColorRange;
 };
 
 export type RgbMappingArrays = {
--- a/src/mapDataManipulation/const.ts
+++ b/src/mapDataManipulation/const.ts
@@ -1,6 +1,9 @@
 export type Effects = {
   gain?: number;
   gamma?: number;
+  redRange?: number[];
+  greenRange?: number[];
+  blueRange?: number[];
 };
 
 export type RgbMappingArrays = {
--- a/src/mapDataManipulation/effectFunctions.ts
+++ b/src/mapDataManipulation/effectFunctions.ts
@@ -2,6 +2,7 @@ import { Effects, RgbMappingArrays } from 'src/mapDataManipulation/const';
 import {
   isEffectSet,
   changeRgbMappingArraysWithFunction,
+  changeRgbMappingArrayInterval,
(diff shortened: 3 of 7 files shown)


src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -24,6 +35,27 @@ export function changeRgbMappingArraysWithFunction(
...
+    let newX = newMin + ((newMax - newMin) / (oldMax - oldMin)) * (x - oldMin);
+    if (strictlyLimitValuesToInterval) {
+      newX = Math.max(newX, newMin);
+      newX = Math.min(newX, newMax);
+    }
+    return newX;
+  };
+
+  rgbMappingArray = changeRgbMappingArrayWithFunction(rgbMappingArray, transformValueToInterval);
+  return rgbMappingArray;
```
sinergise-anze, line 56: `rgbMappingArray` (which comes from params) is being overwritten here. It doesn't matter much in this case, but maybe it is best to avoid it? I would suggest simply returning:
```
return changeRgbMappingArrayWithFunction(rgbMappingArray, transformValueToInterval);
```

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -24,6 +35,27 @@ export function changeRgbMappingArraysWithFunction(
 // [0,255] to [0,1]: a = 0, b = 255; c = 0, d = 1
 // [0,1] to [0,255]: a = 0, b = 1; c = 0, d = 255
 
+export function changeRgbMappingArrayInterval(
+  rgbMappingArray: number[],
+  oldMin: number,
+  oldMax: number,
+  newMin: number,
+  newMax: number,
+  strictlyLimitValuesToInterval: boolean = false,
```
sinergise-anze, line 44: I think the term commonly used is `cap`, so this could be `capValuesToRange`.

Also, there is some inconsistency about the naming - `range` and `interval` are used interchangeably. I would suggest using just one of the terms (imho `range` is more fitting).

Neither of the issues are very important though, as it is clear enough from the context what is happening - fftr either way.

**EDIT:** actually, I think the parameter `strictlyLimitValuesToInterval` is not even needed, as this is something that happens always. Fftr either way.

src/mapDataManipulation/runEffectFunctions.ts
```diff
@@ -29,6 +33,9 @@ export async function runEffectFunctions(originalBlob: Blob, effects: Effects):
   // change the values according to the algorithm (gamma)
   rgbMappingArrays = runGammaEffectFunction(rgbMappingArrays, effects);
 
+  // change the values according to the algorithm (color manipulation)
+  rgbMappingArrays = runSimpleColorEffectFunction(rgbMappingArrays, effects);
```
sinergise-anze, line 37: I think the name `runColorEffectFunction` would be more understandable? Fftr.

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -8,13 +8,24 @@ export function prepareRgbMappingArrays(): RgbMappingArrays {
   };
 }
 
+export function changeRgbMappingArrayWithFunction(
+  rgbMappingArray: number[],
+  transfrormationFunction: Function,
+): number[] {
+  rgbMappingArray = rgbMappingArray.map(x => transfrormationFunction(x));
```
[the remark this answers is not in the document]
zcernigoj: Thanks, fixed.

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -8,13 +8,24 @@ export function prepareRgbMappingArrays(): RgbMappingArrays {
...
 ): RgbMappingArrays {
-  rgbMappingArrays.red = rgbMappingArrays.red.map(x => transfrormationFunction(x));
-  rgbMappingArrays.green = rgbMappingArrays.green.map(x => transfrormationFunction(x));
-  rgbMappingArrays.blue = rgbMappingArrays.blue.map(x => transfrormationFunction(x));
+  rgbMappingArrays.red = changeRgbMappingArrayWithFunction(rgbMappingArrays.red, transfrormationFunction);
+  rgbMappingArrays.green = changeRgbMappingArrayWithFunction(rgbMappingArrays.green, transfrormationFunction);
+  rgbMappingArrays.blue = changeRgbMappingArrayWithFunction(rgbMappingArrays.blue, transfrormationFunction);
+  // rgbMappingArrays.red = rgbMappingArrays.red.map(x => transfrormationFunction(x));
+  // rgbMappingArrays.green = rgbMappingArrays.green.map(x => transfrormationFunction(x));
+  // rgbMappingArrays.blue = rgbMappingArrays.blue.map(x => transfrormationFunction(x));
```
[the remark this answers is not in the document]
zcernigoj: Thanks, that was my goal - to get an opinion on which way is better. I agree. `changeRgbMappingArrayWithFunction` is used in one other function so I will remove that in a next commit.

src/mapDataManipulation/const.ts
```diff
@@ -1,6 +1,9 @@
 export type Effects = {
   gain?: number;
   gamma?: number;
+  redRange?: number[];
+  greenRange?: number[];
+  blueRange?: number[];
```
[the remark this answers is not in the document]
zcernigoj: Thanks, it makes sense, yes. I created a custom type `ColorRange` ([type `Range` already exists in typescript](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.range.html))

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: Thanks, did that, but also removed the usage of `changeRgbMappingArrayWithFunction` and now it's just

```javascript
return rgbMappingArray.map(x => transformValueToInterval(x));
```

src/mapDataManipulation/const.ts
```diff
@@ -1,6 +1,14 @@
+export type ColorRange = {
+  from: number;
+  to: number;
+};
```
sinergise-anze, line 4: This type should probably be exported from the library? (index.js)

zcernigoj: True, and Effects type also. Will export it, thanks.

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: Thanks, true, `capValuesToRange` would be more appropriate. I decided to not use that parameter.

src/mapDataManipulation/const.ts
zcernigoj: Exported both types.

src/mapDataManipulation/runEffectFunctions.ts
zcernigoj: Thanks, I guess so, I don't have a strong opinion because I don't know how the more advanced one will be named or how exactly it will be implemented.

Changed it to `runColorEffectFunction` as suggested.

zcernigoj:
Last thing: adding the documentation to readme.

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -10,38 +10,54 @@ export function prepareRgbMappingArrays(): RgbMappingArrays {
...
+  newMin: number,
+  newMax: number,
+): number[] {
+  const transformValueToRange = (x: number): number => {
+    let newX = newMin + ((newMax - newMin) / (oldMax - oldMin)) * (x - oldMin);
+    newX = Math.max(newX, newMin);
+    newX = Math.min(newX, newMax);
+
+    return newX;
+  };
```
sinergise-anze, line 40: As we have discussed, I would suggest moving this function outside to avoid code duplication. 

It is also possible that you will be left with function(s) with a single statement, in which case I would suggest removing them and performing that statement directly.

Still, this is mostly a cleanup, so fftr.

src/mapDataManipulation/effectFunctions.ts
```diff
@@ -40,3 +41,44 @@ export function runGammaEffectFunction(
...
   return rgbMappingArrays;
 }
+
+export function runColorEffectFunction(
+  rgbMappingArrays: RgbMappingArrays,
+  effects: Effects,
+): RgbMappingArrays {
+  if (!isEffectSet(effects.redRange) && !isEffectSet(effects.greenRange) && !isEffectSet(effects.blueRange)) {
+    return rgbMappingArrays;
+  }
```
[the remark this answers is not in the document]
zcernigoj: Removed. Resolving

README.md
```diff
@@ -171,7 +171,9 @@ It is also possible to determine whether a layer supports a specific ApiType:
 
 When requesting an image, effects can be applied to visually improve the image.
 To apply the effects, the `effects` param in `getMapParams` should be present, containing the desired effects.
-Supported effects are gain and gamma. 
+Supported effects are `gain`, `gamma`, `redRange`, `greenRange` and `blueRange`. 
+All of them accept the values between 0 and 1.
```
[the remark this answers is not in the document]
zcernigoj: True, thanks. Fixed & resolving

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: Thanks, did that. It could probably be cleaned up even further, but maybe do that in a separate MR.

PR #109, opened 2020-06-11
Test/axios interceptors

sinergise-anze:
An important part of sentinelhub-js library is its handling of network errors, which couldn't be tested by mocking Axios because then interceptors are disabled. This MR replaces axios mocking with _network requests_ mocking using [axios-mock-adapter](https://github.com/ctimmerm/axios-mock-adapter).

- a test for retries was added
- existing tests were adapted to the new mocking framework
- also fixed: `LinkType` should be exposed by the library
- default number of retries is lowered to 2 (== 3 tries altogether)

**EDIT:** the default number of retries was previously set to `5`, which means 6 tries altogether - this is a pretty big number, especially as we don't yet have exponential backoff or similar mechanism in place.


src/layer/__tests__/retries.ts
```diff
@@ -0,0 +1,137 @@
...
+    acquisitionMode: AcquisitionMode.IW,
+    polarization: Polarization.DV,
+    resolution: Resolution.HIGH,
+  });
+
+  // we need to adjust jest timeout until we have support for setting the delay when retrying, or
+  // the test will time out:
+  jest.setTimeout(7000);
+
+  mockNetwork.onPost().replyOnce(500);
```
dgostencnik, line 39: Shouldn't this be "binded"  to endpoint used by `findTiles`? I know it doesn't matter in this case, but this code will probably be used as a template for other tests. fftr

src/layer/__tests__/WmsLayer.ts
```diff
@@ -38,14 +41,17 @@ test('WmsLayer.getMapUrl returns an URL', () => {
...
+test('WmsLayer.getMap makes an appropriate request', async () => {
   const bbox = new BBox(CRS_EPSG4326, 19, 20, 20, 21);
   const layerId = 'PROBAV_S1_TOA_333M';
   const layer = new WmsLayer({
     baseUrl: 'https://proba-v-mep.esa.int/applications/geo-viewer/app/geoserver/ows',
     layerId,
   });
 
+  mockNetwork.reset();
+  mockNetwork.onAny().replyOnce(200, ''); // we don't care about response
```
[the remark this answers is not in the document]
sinergise-anze: No, we really don't. ;-) 

This test checks only that the request to the service is correct - it doesn't check anything else. Which means that it doesn't matter what the response is, when the request is made, the test has already either passed or failed. 

Do you think the comment needs to be clarified?

src/layer/__tests__/retries.ts
sinergise-anze: On one hand it would make sense to make sure the network calls are completely correct, but on the other hand the tests should have well-defined scope. In this case the scope is just retrying of network requests - we shouldn't know about the specifics of the call being made. If we are testing URL we should probably also test request parameters...

However, we _are_ using a specific network response and comparing it to known return values (to make sure that the method really succeeded). So I'm not 100% sure what the best approach would be here.

I would suggest we leave this as-is, and refactor later if needed, does that sound OK to you?

src/layer/__tests__/WmsLayer.ts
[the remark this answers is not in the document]
zcernigoj: Talked live, I agree, this format checking doesn't belong here so we really don't care about the response.

Maybe an explanation of why we don't care about the response would be beneficial. I was thinking something like this would do (copied from the change below and modified):
`// we don't care about the response, we will just inspect the request params`

src/layer/__tests__/retries.ts
```diff
@@ -0,0 +1,83 @@
...
+  mockNetwork.reset();
+  mockNetwork.onPost().replyOnce(500);
+  mockNetwork.onPost().replyOnce(500);
+  mockNetwork.onPost().replyOnce(500);
+  mockNetwork.onPost().replyOnce(500);
+
+  await expect(layer.findTiles(bbox, fromTime, toTime, null, null, { retries: 3 })).rejects.toThrow();
+
+  expect(mockNetwork.history.post.length).toBe(4);
+});
```
dgostencnik, line 83: I wanted to add  test for disabled retrying. According to documentation `The default of `retries` is `null` (disabled).` I would expect that making request without  RequestConfiguration  or with  RequestConfiguration {retries:null} or {retries:0} would result in single request being made, but this is not the case.  

```
test("Doesn't retry if retrying is disabled", async () => {
  jest.setTimeout(3 * 3000 + 1000);

  const { fromTime, toTime, bbox, layer } = constructFixtureFindTiles({});
  mockNetwork.reset();
  mockNetwork.onPost().replyOnce(500);
  mockNetwork.onPost().replyOnce(500);
  mockNetwork.onPost().replyOnce(500);

  await expect(layer.findTiles(bbox, fromTime, toTime, null, null, { retries: null })).rejects.toThrow();
});
```

src/layer/__tests__/retries.ts
```diff
@@ -0,0 +1,83 @@
...
+  mockNetwork.onPost().replyOnce(500);
+  mockNetwork.onPost().replyOnce(500);
+
+  await expect(layer.findTiles(bbox, fromTime, toTime)).rejects.toThrow();
+
+  expect(mockNetwork.history.post.length).toBe(3);
+});
+
+test('Fails if max. retries (explicitly set) was exceeded', async () => {
+  jest.setTimeout(3 * 3000 + 1000);
```
dgostencnik, line 71: This waiting is really annoying.

src/layer/__tests__/retries.ts
sinergise-anze: Great catch, thank you!

There was a mismatch between the docs and the implementation, and the implementation didn't work anyway (no way to disable retries). I have now:
- set default value to `2` (sentinelhub-py has a default value for `max_download_attempts` set to 4, but that seemed a bit high to me for our use-case - thank you @zcernigoj for the idea of checking there!),
- made `null` a "use default" value, similar to how it is used with parameters
- fixed the documentation and tests to reflect that.

Does this change make sense or do you think some other values would be more appropriate?

src/layer/__tests__/WmsLayer.ts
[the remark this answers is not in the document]
sinergise-anze: Thank you, fixed!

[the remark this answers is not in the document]
zcernigoj: Perfect! Resolving.

PR #110, opened 2020-06-23
Fix/auth token in reqs config

sinergise-anze:
Sometimes apps manage more than a single auth token at a time, in which case having a single global setting is cumbersome to use. This PR adds an `authToken` setting to `RequestsConfig`, which is then used instead of global auth token.

Additional changes:
- updated README and added relevant tests
- fixed table of contents in README
- `setAuthToken` now allows `null` as token (to allow resetting of auth token)
- `overrideConstructorParams` param of `LayersFactory.makeLayers()` is explicitly allowed to be `null` (we expected `null` even before, just the definition was missing)

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,7 +349,7 @@ Getting the authentication token by calling `requestAuthToken()` with the OAuth 
   const after = isAuthTokenSet(); // true
 ```
 
-Alternatively, authentication token can be set on a per-request basis, which also overrides any global (`setAuthToken()`) token:
+Alternatively, authentication token can be set on a per-request basis, which also overrides any global token that was set by `setAuthToken`:
 
 ```javascript
   const requestsConfig = {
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -350,7 +343,19 @@ Getting the authentication token by calling `requestAuthToken()` with the OAuth
   const clientId = /* OAuth Client's id, best to put it in .env file and use it from there */;
   const clientSecret = /* OAuth client's secret, best to put it in .env file and use it from there */;
   const authToken = await requestAuthToken(clientId, clientSecret);
+
+  const before = isAuthTokenSet(); // false
   setAuthToken(authToken);
+  const after = isAuthTokenSet(); // true
+```
+
+Alternatively, authentication token can be set on a per-request basis, which also overrides any global (`setAuthToken()`) token:
```
zcernigoj, line 352: I would write that differently - **... overrides any global token that was set by / with `setAuthToken()`**

sinergise-anze: Better, thank you! Fixed.

zcernigoj: Nice. Resolving

PR #112, opened 2020-07-07
handle undefined reponseType as json, axios does this by default

guss84:
A few functions do not have `responseType` set in the axios `requestConfig`, which causes these requests to not be cached.
Since axios handles `undefined` `responseType` as `json`, we could take a similar approach to this and not add `responseType` to each request.


sinergise-anze:
This looks good. We discussed adding tests live, so I won't add :+1: just yet - please assign back when you feel it is ready to be reviewed.

sinergise-anze:
As discussed, we will be adding tests in a separate MR (incoming).

PR #113, opened 2020-07-08
Feature/expose cache

guss84:
* Changed `useCache` to `expiresIn` which is in seconds
* Expose `expiresIn` to `reqConfig`
* Changed the key hashing algorithm, as the previous one required some dependencies on nodejs


src/layer/__tests__/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+const mockNetwork = new MockAdapter(axios);
+
+describe('Testing caching', () => {
+  beforeEach(() => {
+    Object.assign(global, makeServiceWorkerEnv());
+    jest.resetModules();
+  });
+
+  it('should fetch a request and cache it, where 2nd request is served from the cache', async () => {
+    jest.setTimeout(7000);
```
sinergise-anze, line 17: Since we are not timing out on requests, this line should probably be removed (it is not needed and indicates that this is a long-running test, which it is not).

package.json
```diff
@@ -7,11 +7,13 @@
   "dependencies": {
     "@turf/area": "^6.0.1",
     "@turf/helpers": "^6.1.4",
+    "@types/service-worker-mock": "^2.0.1",
     "@types/xml2js": "^0.4.4",
     "axios": "^0.18.1",
     "moment": "^2.24.0",
     "polygon-clipping": "^0.14.3",
     "query-string": "^6.4.2",
+    "service-worker-mock": "^2.0.5",
```
sinergise-anze, line 16: Since this is only used by tests, should we add it to `devDependencies`?

package.json
```diff
@@ -65,7 +67,8 @@
     "testPathIgnorePatterns": [
       "/node_modules/",
       "/dist/",
-      "[.]fixtures[.]ts"
+      "[.]fixtures[.]ts",
+      "fixtures[.].*[.]ts"
```
sinergise-anze, line 71: As discussed, we should probably use:
```
"^fixtures[.].*[.]ts"
```
and rename existing fixtures to `fixtures.*.ts`.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -147,7 +146,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
 
   protected async fetchEvalscriptUrlIfNeeded(): Promise<void> {
     if (this.evalscriptUrl && !this.evalscript) {
-      const response = await axios.get(this.evalscriptUrl, { responseType: 'text', useCache: true });
+      const response = await axios.get(this.evalscriptUrl, { responseType: 'text' });
```
sinergise-anze, line 149: I know this is out of scope for this MR, but there is a bug here - `fetchEvalscriptUrlIfNeeded` does not use `reqConfig`. 

Because of this, removing `useCache` now changes the behaviour of the method. To fix this, `expiresIn` should be added. But maybe it is easier to just fix the bug?

src/layer/__tests__/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    const { fromTime, toTime, bbox, layer, mockedResponse } = constructFixtureFindTiles({});
+    const requestsConfig = {
+      expiresIn: 1,
+    };
+    mockNetwork.reset();
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+
+    await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+    await new Promise(r => setTimeout(r, 2000));
```
sinergise-anze, line 56: Nitpick: `1100` should be more than enough... :) Fftr.

README.md
```diff
@@ -291,6 +291,14 @@ try {
 }
 ```
 
+To enable caching for the requests, one can add `expiresIn` to the requests configuration object. The values are in seconds.
+```
+// cache is valid for 30 minutes
+const requestsConfig = {
+  expiresIn: 1800,
```
sinergise-anze, line 298: Looking at this interface description, it is not clear the `expiresIn` enables caching (or that it is connected to cache at all, for that matter).

Maybe we should change the interface so that `cache` key must be set for caching to be enabled?

Also, maybe we should rename `expiresIn` to `cacheExpiresIn`, `cacheLifetime` or similar? And set the default value for it when caching is enabled?

Note that this would be a breaking change if we were to change it later, so it pays to put some effort in designing a better interface now...

src/layer/__tests__/cache.ts
guss84: I have removed all the `jest.setTimeout`

package.json
guss84: Yes thanks, moved this to `devDependencies`

src/layer/__tests__/cache.ts
guss84: True, changed to `1100`

README.md
guss84: The interface has changed to following and future elements like `target` will also move in to the object
```
  cache: {
  expiresIn: 1800,
  }
};
```

src/layer/AbstractSentinelHubV3Layer.ts
guss84: It was not a bug, I thought it would be best if consumers handle the caching, but this would be a breaking change.
The default behavior has been added back, and consumers can always disable caching on requests.

README.md
```diff
@@ -291,6 +291,25 @@ try {
 }
 ```
 
+To enable caching for the requests, one can add `expiresIn` to the requests configuration object. The values are in seconds.
+```
+// cache is valid for 30 minutes
+const requestsConfig = {
+  cache: {
+  expiresIn: 1800,
```
sinergise-anze, line 299: Typo: indentation is wrong.

EDIT: below also.

README.md
```diff
@@ -291,6 +291,25 @@ try {
...
+  }
+};
+```
+
+To disbale caching for the requests that are enbaled by default, you can assign 0 to expiresIn.
```
sinergise-anze, line 304: Typos: `disbale`, `enbaled`.

I would also suggest putting "expiresIn" in ticks (`expiresIn` instead of expiresIn).

src/layer/AbstractSentinelHubV3Layer.ts
sinergise-anze: I don't think we are talking about the same thing? I think the bug is in that user can't affect `axios.get` through `reqConfig`. 

Note that this is not a new bug. The reverting of default caching values now means that there is no change because of this MR, so if you feel this is out of scope, feel free to resolve.

It still needs to be fixed though... let me know if you resolve and I can add a task for this.

src/layer/__tests__/cache.ts
```diff
@@ -0,0 +1,115 @@
...
+  it('test that no responses are cached', async () => {
+    const { fromTime, toTime, bbox, layer, mockedResponse } = constructFixtureFindTiles({});
+    mockNetwork.reset();
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+
+    await layer.findTiles(bbox, fromTime, toTime);
+    await layer.findTiles(bbox, fromTime, toTime);
+
+    expect(mockNetwork.history.post.length).toBe(2);
```
sinergise-anze, line 77: Since caching is set to half an hour by default, I would expect this test to fail. What am I missing?

src/utils/cacheHandlers.ts
```diff
@@ -0,0 +1,190 @@
...
+      cache.delete(key);
+    }
+  });
+};
+
+const isRequestCachable = (request: AxiosRequestConfig): boolean => {
+  if (!(request && request.cache && request.cache.expiresIn)) {
+    return false;
+  }
+  // cache can be disbabled with expiresIn: 0;
```
sinergise-anze, line 184: Typo: disbabled

src/layer/__tests__/cache.ts
guss84: We never cached `findTiles`, and since `cache` is not set in this test there should be no cache

src/layer/AbstractSentinelHubV3Layer.ts
guss84: Default caching has been added back, and users can override this behavior with `reqConfig`

README.md
guss84: Fixed

README.md
guss84: Fixed

src/utils/cacheHandlers.ts
guss84: Fixed

src/layer/AbstractSentinelHubV3Layer.ts
guss84: Do you think `fetchEvalscriptUrlIfNeeded` should use the same `reqConfig` as `getMap`?

src/utils/cacheHandlers.ts
```diff
@@ -0,0 +1,190 @@
...
+export const DEFAULT_CACHE_CONFIG = { expiresIn: 1800 };
+const SENTINEL_HUB_CACHE = 'sentinelhub-v1';
+const EXPIRY_HEADER_KEY = 'Cache_Expires';
+
+export type CacheConfig = {
+  expiresIn: number;
+};
+
+export const fetchCachedResponse = async (request: any): Promise<any> => {
+  if (!isRequestCachable) {
```
sinergise-anze, line 13: This check is always true.

guss84: Thanks, missed this one. It has been fixed

src/layer/AbstractSentinelHubV3Layer.ts
guss84: This has been fixed by passing `innerReqConfig` to `fetchEvalscriptUrlIfNeeded`

src/layer/__tests__/cache.ts
guss84: As discussed ,  `cacheConfig` is passed to `getAxiosReqParams` in order to make it more explicit what is cached by default and what is not.

sinergise-anze: Nice!

PR #118, opened 2020-07-13
Feature/huge maps

sinergise-anze:
- added `getHugeMap` which allows us to seamlessly request images which are bigger than what service provides (useful for high-res print image download and for pins story builder, both in EO Browser)
- exposed `canvasToBlob` and `drawBlobOnCanvas` functions (needed by `getHugeMap`; these methods are duplicated in numerous other places in EO Browser and other apps - hoping this to be the implementation that will be used everywhere)

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,29 @@ Alternatively, authentication token can be set on a per-request basis, which als
   const img = await layer.getMap(getMapParams, ApiType.PROCESSING, requestsConfig);
 ```
 
+## Utility functions
+
+### Async conversion between Blob and Canvas
+
+Function `drawBlobOnCanvas` allows drawing a Blob on existing canvas element:
+
+```javascript
+  const blob = await layer.getMap(params, ApiType.WMS);
+
+  const canvas = document.createElement('canvas');
+  canvas.width = params.width;
+  canvas.height = params.height;
+  const ctx = canvas.getContext('2d');
+
+  await drawBlobOnCanvas(ctx, blob, 0, 0);
+```
+
+Function `canvasToBlob` converts an existing canvas to Blob:
+
+```javascript
+  const blob = await canvasToBlob(canvas);
+```
+
 ## Debugging
 
 This library is an abstraction layer that provides nice interface for accessing the underlying services, which simplifies development - but when requests fail, it is sometimes difficult to understand why. To enable easier debugging, `setDebugEnabled` can be used:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -195,6 +195,16 @@ When retrieving an image URL (via `getMapUrl()`) with effects applied, an error
   const imageBlob2 = await layer.getMap(getMapParamsWithEffects, ApiType.PROCESSING);
 ```
 
+### Stitching images
+
+Services limit the size of the output image per request (5000px for OGC and 2500px for Processing API). If we need a bigger image, we can issue multiple requests and stitch the results together in a canvas. A utility method `getHugeMap` allows us to do that seamlessly.
```
zcernigoj, line 200: I think both OGC (WMS) and Processing API are limited to 2500px. Will talk live to give more info.

README.md
```diff
@@ -358,6 +368,29 @@ Alternatively, authentication token can be set on a per-request basis, which als
...
+
+  const canvas = document.createElement('canvas');
+  canvas.width = params.width;
+  canvas.height = params.height;
+  const ctx = canvas.getContext('2d');
+
+  await drawBlobOnCanvas(ctx, blob, 0, 0);
+```
+
+Function `canvasToBlob` converts an existing canvas to `Blob`:
```
zcernigoj, line 388: I would change this text to:

Function `canvasToBlob`  converts the provided canvas to `Blob`:


(since it accepts the `canvas` as a parameter)

src/utils/canvas.ts
```diff
@@ -0,0 +1,26 @@
+import { MimeType } from 'src/layer/const';
+
+export async function drawBlobOnCanvas(
```
zcernigoj, line 3: Would it be better for `drawBlobOnCanvas` to get the `canvas`, not just its `context`? (would be more in line with `canvasToBlob()`)

sinergise-anze: At least how it is used in `getHugeMap`, this is more convenient because we can create one huge canvas and then draw the blobs on it. We can however later add another function (`blobToCanvas`?) if needed, if you agree?

README.md
sinergise-anze: Thank you, your version is better, fixed.

README.md
sinergise-anze: Thank you, fixed!

src/utils/canvas.ts
zcernigoj: I agree. Resolving

README.md
zcernigoj: Nice, resolving.

README.md
zcernigoj: Resolving.

PR #120, opened 2020-07-14
Fix / legacyGetMap... functions additional params

zcernigoj:
Adding support for params in `legacyGetMapFromParams`, `legacyGetMapFromUrl` which are not supported in `wmsParams`.

Currently divided to 2 params:
- `overrideLayerConstructorParams`: params that are used when constructing the `Layer` (or changing the `Layer`'s property)
- `overrideGetMapParams`: params that are supported in `GetMapParams` for `Layer.getMap()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,11 +379,14 @@ If we already have a WMS GetMap URL, we can use it directly:
   const imageBlob4 = await legacyGetMapFromUrl(fullUrlWithWmsQueryString, ApiType.PROCESSING);
 ```
 
-Gain and gamma effects are also supported in these two functions as a part of `wmsParams` or `fullUrlWithWmsQueryString`.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -325,6 +325,21 @@ If we already have a WMS GetMap URL, we can use it directly:
 
 Gain and gamma effects are also supported in these two functions as a part of `wmsParams` or `fullUrlWithWmsQueryString`.
 
+`legacyGetMapFromParams` and `legacyGetMapFromUrl` also accept the parameters that are not supported in WMS GetMap URL but are supported.
```
Ardweaden, line 328: Nitpick, but but this would be a lot clearer of it said smth like `legacyGetMapFromParams` and `legacyGetMapFromUrl` also accept the parameters that are not supported in WMS GetMap standard, but are supported by Sentinel Hub/sentinelhub-js.` or smth.

zcernigoj: Haha forgot to actually finish the sentence 😅 Thanks for spotting. And the suggestion also makes sense, will change.

zcernigoj: Thanks again. I paid more attention to that section of readme and changed the text a bit more.

PR #123, opened 2020-07-17
Feature/cache targets

guss84:
* Add support for cache targets in `reqConfig`

Can specify a list of targets that are ordered by priority for checking if the target is available, where caching will always fallback to in memory caching. By default the library will try to use cache-api.
More targets could be added in a separate MR.

### DIFF
--- a/src/layer/__tests__/cache.ts
+++ b/src/layer/__tests__/cache.ts
@@ -208,3 +208,36 @@ describe('Testing cache targets', () => {
     expect(fromCacheResponse.tiles).toStrictEqual(expectedResultTiles);
   });
 });
+
+describe('Testing cache targets when cache_api is not available', () => {
+  beforeEach(() => {
+    Object.assign(global, fetch);
+    jest.resetModules();
+    memoryCache.clear();
+  });
+
+  it('should default to memory if window.caches is undefined', async () => {
+    const { fromTime, toTime, bbox, layer, mockedResponse, expectedResultTiles } = constructFixtureFindTiles(
+      {},
+    );
+    const requestsConfig = {
+      cache: {
+        expiresIn: 60,
+      },
+    };
+    mockNetwork.reset();
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+
+    const responseFromMockNetwork = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+    const fromCacheResponse = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+    const cacheFromMemoryItem = memoryCache.has(mockNetwork.history.post[0].cacheKey);
+
+    expect(self.caches).toBeUndefined();
+    expect(cacheFromMemoryItem).toBeTruthy();
+    expect(cacheFromMemoryItem).not.toEqual(null);
+    expect(mockNetwork.history.post.length).toBe(1);
+    expect(responseFromMockNetwork.tiles).toStrictEqual(expectedResultTiles);
+    expect(fromCacheResponse.tiles).toStrictEqual(expectedResultTiles);
+  });
+});
--- a/src/layer/__tests__/cache.ts
+++ b/src/layer/__tests__/cache.ts
@@ -8,6 +8,7 @@ import { constructFixtureFindTiles } from './fixtures.findTiles';
 import { constructFixtureGetMap } from './fixtures.getMap';
 import { ApiType } from 'src';
 import { setAuthToken } from 'src/auth';
+import { memoryCache, CacheTarget, CACHE_API_KEY } from 'src/utils/Cache';
 
 const mockNetwork = new MockAdapter(axios);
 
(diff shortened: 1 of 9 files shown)


src/layer/__tests__/cache.ts
```diff
@@ -113,3 +113,188 @@ describe('Testing caching', () => {
     expect(mockNetwork.history.post.length).toBe(2);
   });
 });
+
+describe('Testing cache targets', () => {
+  beforeEach(() => {
+    Object.assign(global, makeServiceWorkerEnv(), fetch);
+    jest.resetModules();
```
sinergise-anze, line 120: The meaning of these two lines is not apparent to me... If they are needed, could you maybe add a comment explining their purpose?

src/layer/__tests__/cache.ts
```diff
@@ -113,3 +113,188 @@ describe('Testing caching', () => {
...
+    memoryCache.clear();
```
sinergise-anze, line 121: For consistency, should we use `invalidateCaches` before each test as it is done in the other test files? I assume other caches should be cleared too, not just memory?

src/layer/__tests__/cache.ts
```diff
@@ -113,3 +113,188 @@ describe('Testing caching', () => {
...
+        targets: [CacheTarget.CACHE_API],
+      },
+    };
+    mockNetwork.reset();
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+    mockNetwork.onPost().replyOnce(200, mockedResponse);
+
+    const responseFromMockNetwork = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+    const fromCacheResponse = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+
```
sinergise-anze, line 140: I would suggest adding another assert here:
```
```

Actually, I think the tests would be much stricter (and more readable) if every `findTiles` call would be followed with an `expect`, if you agree?

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+  MEMORY = 'MEMORY',
+}
+export type CacheResponse = {
+  data: any;
+  status: number;
+  statusText: string;
+  headers: any;
+};
+
+export const CACHE_API_KEY = 'sentinelhub-v1';
```
sinergise-anze, line 15: Nitpick: maybe this should be `sentinelhub-js-v1`?

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+export const SUPPORTED_TARGETS = [CacheTarget.CACHE_API, CacheTarget.MEMORY];
+const FALLBACK_TARGET = CacheTarget.MEMORY;
+
+export const memoryCache = new Map();
```
sinergise-anze, line 19: If we are using `invalidateCaches` in tests, this can be made internal.

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+// By default we will use cache_api if availble, if not we will cache in memory
+// user can also specify to just use in memory caching
+// If user provides a CacheTarget.CACHE_API and cache_api is not availble we will fallback to memory
+export function cacheFactory(optionalTargets: CacheTargets): ShCache {
+  const targets = optionalTargets || SUPPORTED_TARGETS;
+  const target = getFirstUsuableTarget(targets);
+  return constructCache(target);
+}
+
+function getFirstUsuableTarget(targets: CacheTargets): CacheTarget {
```
sinergise-anze, line 32: Typo: `Usuable`

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+  let firstTargetToUse = FALLBACK_TARGET; // default to memory if target is not supported
```
sinergise-anze, line 33: Maybe a better idea would be to default to _no cache_ instead of memory? If developer explicitly set just `[CACHE_API]`, they would probably assume that memory would not be used as target?

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+  keys(): Promise<string[]>;
+  delete(key: string): Promise<void>;
+  getHeaders(key: string): Promise<Record<string, any>>;
+  invalidate(): void;
+}
+
+class MemoryCache implements ShCache {
+  private cache: Record<string, any>;
+  public constructor() {
+    this.cache = memoryCache;
```
sinergise-anze, line 78: Since we are always using an "external" (to class) resource to save cache data (`memoryCache`), and since we can't change `this.cache` anyway, I would suggest removing `this.cache` var and simply using `memoryCache` directly:
```
  public async set(key: string, response: AxiosResponse): Promise<void> {
    memoryCache.set(key, response);
  }
```

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  public async set(key: string, response: AxiosResponse): Promise<void> {
+    this.cache.set(key, response);
```
sinergise-anze, line 82: Does this work? We are setting a `Response` here. When we retrieve it via `get`, we also read its `.data`. IIRC we can do this only once, so the next time we try to call `get` it would fail?

It might be a good idea to save an object instead of `Response`:
```
  this.cache.set(key, {
      data: cachedResponse.data,
      status: cachedResponse.status,
      statusText: cachedResponse.statusText,
      headers: cachedResponse.headers,
  });
```

A separate note: this scenario is a good candidate for testing ("what happens if I read the cached entry two times" - for memoy and for Cache API separately).

src/utils/Cache.ts
```diff
@@ -0,0 +1,246 @@
...
+    const response: Response = await cache.match(key);
+    if (!response) {
+      return null;
+    }
+
+    return {
+      data: await this.deSerializeResponseData(response, responseType),
+      status: response.status,
+      statusText: response.statusText,
+      headers: await this.deSeriarialeHeaders(response.headers),
```
sinergise-anze, line 146: Typo: `deSeriarialeHeaders`

src/utils/Cache.ts
guss84: We could, but then we should also delete 'sentinelhub-v1'  from users' browsers?

src/utils/Cache.ts
guss84: Cache api requires a `Response` where as memory caching can be any javascript object, so here we are using the `AxiosResponse` directly as value, so in other words we do not have the issue of using a `Response` body twice when using the cache from memory

src/utils/Cache.ts
sinergise-anze: Ah. Nevermind then, resolving. :)

src/utils/Cache.ts
guss84: I will add another test, just to cover our bases

src/utils/Cache.ts
Shooshte: Alternatively if we are exposing it the function should be documented properly with an example. Don't really know who should decide this. I would say expose it and document it if the users wants to invalidate the cache for some reason (expose more control over caching to the consumer).

src/utils/Cache.ts
Shooshte: Resolving since defaulting to no cache was added.

src/utils/cacheHandlers.ts
```diff
@@ -102,36 +63,16 @@ export const saveCacheResponse = async (response: AxiosResponse): Promise<any> =
...
-      throw new Error('Unsupported response type: ' + request.responseType);
-  }
-  cache.put(response.config.cacheKey, new Response(responseData, response));
+  shCache.set(response.config.cacheKey, response);
   return response;
 };
 
-const cacheStillValid = (response: Response): boolean => {
-  if (!response) {
+const cacheStillValid = (headers: Record<string, any>): boolean => {
```
Shooshte, line 75: I think we should add a unit test for this as well just to ensure that we do not introduce any breaking changes in the future.

src/layer/__tests__/cache.ts
guss84: Done

src/utils/Cache.ts
guss84: `memoryCache` is not exported anymore as we are using `invalidateCaches` in the tests now

src/utils/Cache.ts
guss84: Fixed

src/utils/Cache.ts
guss84: This has been changed so that we do not fallback to memory cache. If the user doesnt supply targets then the default targets will be used and will sill fallback to memory if the other targets are undefined

src/utils/Cache.ts
guss84: Fixed

src/layer/__tests__/cache.ts
guss84: I have restructured the tests a bit per your suggestion

src/layer/__tests__/cache.ts
guss84: I have removed `jest.resetModules` and added a comment to the remaining lines

src/utils/cacheHandlers.ts
guss84: Wouldnt hurt, added a unit test for the function `cacheStillValid`

src/utils/Cache.ts
guss84: Removed `this.cache` from `memoryCache` and `CacheApi`. Once it was removed from `CacheApi` another issue arose and that was that the return from `keys()` was incorrect, so this has also been fixed

PR #125, opened 2020-07-23
fix path to declaration

guss84:
* Fix path to the type declaration file

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,9 +68,9 @@
       "[.]fixtures[.]ts"
     ]
   },
-  "types": "dist/index.d.ts",
+  "types": "dist/src/index.d.ts",
   "files": [
     "dist",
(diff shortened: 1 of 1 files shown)


guss84:
> Should it be investigated why `/dist/index.d.ts` doesn't have all the definitions or was it the wrong file from the start?
> 
> I think the change is ok though.

`/dist/index.d.ts` never existed

zcernigoj:
> > Should it be investigated why `/dist/index.d.ts` doesn't have all the definitions or was it the wrong file from the start?
> > I think the change is ok though.
> 
> `/dist/index.d.ts` never existed

Ah thanks, it turned out I had some ancient old dist folder, that's why `dist/index.d.ts` file existed on my computer.

PR #127, opened 2020-07-28
legacyGetMapFromParams fixes for effects

zcernigoj:
- overrideLayerConstructorParams was missing in the fallback path of the flow in `legacyGetMapFromParams`
- effects were not handled correctly in `legacyGetMapFromParams`

### DIFF
--- a/src/legacyCompat.ts
+++ b/src/legacyCompat.ts
@@ -89,6 +89,21 @@ export async function legacyGetMapFromParams(
 
   const updatedGetMapParams = { ...getMapParams, ...overrideGetMapParams };
 
+  let combinedEffects = {};
+  if (getMapParams.effects && Object.keys(getMapParams.effects).length > 0) {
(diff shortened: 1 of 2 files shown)


src/legacyCompat.ts
```diff
@@ -89,6 +89,21 @@ export async function legacyGetMapFromParams(
 
   const updatedGetMapParams = { ...getMapParams, ...overrideGetMapParams };
 
+  let combinedEffects = {};
+  if (getMapParams.effects && Object.keys(getMapParams.effects).length > 0) {
+    combinedEffects = { ...combinedEffects, ...getMapParams.effects };
```
Ardweaden, line 94: Very minor, but I think there is no need for `...combinedEffects` here.

zcernigoj: I know, I added it deliberately ... something between paranoia (something not working again because of some stupid mistake) and "just in case I will need to change something in the near future".

PR #128, opened 2020-08-10
Problem with cacheApi

dgostencnik:
[trello](https://trello.com/c/ZS3gSYlX/66-caching-problem)

It turns out that it is not sufficient to check if `window.caches` object exists, but  some request should be made to determine whether cacheApi can be used or not. 
I was able to reproduce this only in firefox private mode.

### DIFF
--- a/src/utils/Cache.ts
+++ b/src/utils/Cache.ts
@@ -22,16 +22,16 @@ const memoryCache = new Map();
 // By default we will use cache_api if availble, if not we will cache in memory
 // user can also specify to just use in memory caching
 // If user provides a CacheTarget.CACHE_API and cache_api is not availble we will fallback to memory
-export function cacheFactory(optionalTargets: CacheTargets): ShCache {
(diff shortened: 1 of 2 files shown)


src/utils/Cache.ts
```diff
@@ -22,27 +22,42 @@ const memoryCache = new Map();
...
+  if (!window.caches) {
+    return false;
+  }
+  try {
+    const caches = window.caches;
+    //make request to cacheApi to check if it is available
+    await caches.keys();
+    return true;
+  } catch (err) {
+    console.error('CacheApi is not available', err);
```
sinergise-anze, line 52: Maybe a warning would be enough here? It is not really an error, it is working as intended.

Also, it would be nice if it was displayed only once, but that means that this function should remember its result (which might be a good idea from performance point of view too), so I'm not sure if this is still within the scope. 

Fftr and merge.

PR #131, opened 2020-08-11
Fix/caching expiry

guss84:
* Resolves a bug where expired cached items would not be updated when a new network request was made, where these items would stay in cache but would not be used as the expiry stayed the same.
* Bug was seen with getCapabilities so tests for getCapabilities were added along with adding tests to other methods that already had tests.


zcernigoj:
`cacheStillValid()` has this code:

```javascript
if (!headers) {	
  return true;
}
```

Is this assumption correct? I would say that if no headers are provided, you can't check if the `cache` is valid, so I would intuitively return `false`.

FFTR if I'm wrong.

guss84:
> `cacheStillValid()` has this code:
> 
> ```js
> if (!headers) {	
>   return true;
> }
> ```
> 
> Is this assumption correct? I would say that if no headers are provided, you can't check if the `cache` is valid, so I would intuitively return `false` (assuming the default "state" should be `cache` is not valid).
> 
> FFTR if I'm wrong.

I cant think of a reason for this and will have to inspect a bit further to see if there was a valid reason for returning true. If I find a valid reason for it then I will add a comment to that line.

src/layer/__tests__/cache.ts
```diff
@@ -432,3 +486,62 @@ describe('Unit test for cacheStillValid', () => {
     expect(isValid).toBeFalsy();
   });
 });
+
+describe('test getCapabilities', () => {
```
sinergise-anze, line 490: Do we need these tests? Unless I'm missing something, I think their functionality (with regards to caching) is replicated elsewhere... or is the difference in that we are caching GET requests instead of POST?

Also, note that `fetchGetCapabilitiesJson` is not a public method (not sure if it matters in this case).

src/layer/__tests__/cache.ts
```diff
@@ -58,9 +68,26 @@ describe('Testing caching', () => {
...
+      null,
+      null,
+      requestsConfig,
+    );
+    const fromCacheResponse2 = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
+    expect(fromCacheResponse2.tiles).toStrictEqual(expectedResultTiles);
+    expect(fromCacheResponse2.hasMore).toBe(expectedResultHasMore);
+    expect(fromCacheResponse2).toStrictEqual(responseFromMockNetwork2);
 
     expect(mockNetwork.history.post.length).toBe(2);
```
sinergise-anze, line 92: I would suggest being more explicit about which method calls are using network and which are not. In other words, I think we should change these lines to:

```javascript
    const responseFromMockNetwork = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);

    const fromCacheResponse = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
    expect(mockNetwork.history.post.length).toBe(1); // no network request - cache was used
    expect(fromCacheResponse.tiles).toStrictEqual(expectedResultTiles);
    expect(fromCacheResponse.hasMore).toBe(expectedResultHasMore);
    expect(fromCacheResponse).toStrictEqual(responseFromMockNetwork);

    await new Promise(r => setTimeout(r, 1100));

    const responseFromMockNetwork2 = await layer.findTiles(
      bbox,
      fromTime,
      toTime,
      null,
      null,
      requestsConfig,
    );
    expect(mockNetwork.history.post.length).toBe(2);

    const fromCacheResponse2 = await layer.findTiles(bbox, fromTime, toTime, null, null, requestsConfig);
    expect(mockNetwork.history.post.length).toBe(2); // no network request - cache was used
    expect(fromCacheResponse2.tiles).toStrictEqual(expectedResultTiles);
    expect(fromCacheResponse2.hasMore).toBe(expectedResultHasMore);
    expect(fromCacheResponse2).toStrictEqual(responseFromMockNetwork2);
```

src/layer/__tests__/cache.ts
guss84: These tests are not needed, but they are the only tests that are testing cache and using GET requests. I would say there are better methods to test GET requests and caching, so I will remove these tests

src/layer/__tests__/cache.ts
guss84: Changed to what to you suggested

PR #132, opened 2020-08-20
add support for custom effect

zcernigoj:
This mr adds a support to use custom function to change the pixel values. It will be used by EO Browser's advanced rgb effects.

For the sake of being more general (not limiting to the RGB values 0-255) I decided to work with values between 0 and 1.


README.md
```diff
@@ -188,13 +199,21 @@ When retrieving an image URL (via `getMapUrl()`) with effects applied, an error
       redRange: {from: 0.2, to: 0.8},
       greenRange: {from: 0.2, to: 0.8},
       blueRange: {from: 0.2, to: 0.8},
+      customEffect: {
+        redFunction: pixelValue => exampleFunction[pixelValue],
+        greenFunction: pixelValue => exampleFunction[pixelValue],
+        blueFunction: pixelValue => exampleFunction[pixelValue],
+      }
```
sinergise-anze, line 206: I think the interface would be more intuitive (and powerful) if there was only a single function available which could modify all three color bands. Sth. like this:
```
customEffect: ({r, g, b}) => exampleFunction(r, g, b),
```

If you decide to do so, an interesting (and short) `exampleFunction` could be one that rotates the colors:
```
const rotateColors = ({r, g, b}) => ({g, b, r});
```

sinergise-anze (commented):
Since the first comment means quite a few changes, I will stop the code review for now and will get back to it later.

README.md
zcernigoj: Thanks, I changed it to your suggestion. Don't really know why I went with separated functions the first time

sinergise-anze: Looks good, except documentation still needs to be updated in lines 202-206 (and I would suggest using a simpler `exampleFunction`).

src/mapDataManipulation/effectFunctions.ts
```diff
@@ -66,3 +66,28 @@ export function runColorEffectFunction(
...
+
+export function runCustomEffectFunctions(
+  rgbMappingArrays: RgbMappingArrays,
+  effects: Effects,
+): RgbMappingArrays {
+  if (!isEffectSet(effects.customEffect)) {
+    return rgbMappingArrays;
+  }
+
+  let newRgbMappingArrays = { ...rgbMappingArrays };
```
sinergise-anze, line 78: IIUC, this is not needed? It might also be a bit misleading, because even if we are copying the parent object, the values are actually being changed in-place. I would suggest simply using `rgbMappingArrays` directly, to explicitly show that we are mutating the structure.

stories/s2l2a.stories.js
```diff
@@ -843,6 +843,93 @@ export const getMapProcessingGainGammaCheckTransparency = () => {
...
+      },
+    };
+
+    const getMapParamsRGB2 = {
+      ...getMapParams,
+      effects: {
+        customEffect: ({ r, g, b }) => ({
+          r: r < 0.4 ? r * 2 : r,
+          g: g < 0.4 ? g * 2 : g,
+          b: b < 0.4 ? b * 2 : b,
```
sinergise-anze, line 910: I would suggest using a function which more clearly demonstrates the functionality by giving more predictable results, for example:
```
        customEffect: ({ r, g, b }) => ({
          r: r + g + b < 0.6 ? 0 : 1,
          g: r + g + b < 0.6 ? 0 : 1,
          b: r + g + b < 0.6 ? 0 : 1,
        }),
```

src/mapDataManipulation/effectFunctions.ts
```diff
@@ -66,3 +66,28 @@ export function runColorEffectFunction(
 
   return rgbMappingArrays;
 }
+
+export function runCustomEffectFunctions(
```
sinergise-anze, line 70: When playing around with different `customEffect` functions in the storybook example, I noticed that this function gives unexpected results:
```
        customEffect: ({ r, g, b }) => ({
          r: r + g + b < 0.6 ? 0 : 1,
          g: r + g + b < 0.6 ? 0 : 1,
          b: r + g + b < 0.6 ? 0 : 1,
        }),
```

The resulting pixels should be either black or white (unless I'm misunderstanding something), but I get:
image

Since it is difficult to know if the implementation is correct just by looking at the code, I would suggest writing unit tests for `runCustomEffectFunctions` (and other effect functions, though that might be out of scope in this MR).

stories/s2l2a.stories.js
```diff
@@ -843,6 +843,93 @@ export const getMapProcessingGainGammaCheckTransparency = () => {
...
+  imgRGB1.width = '256';
+  imgRGB1.height = '256';
+
+  const imgRGB2 = document.createElement('img');
+  imgRGB2.width = '256';
+  imgRGB2.height = '256';
+
+  const wrapperEl = document.createElement('div');
+  wrapperEl.innerHTML = '<h2>S2L2A getMapProcessingAdvancedRGB</h2>';
+  wrapperEl.innerHTML += '<h4>no effects | same value | function</h4>';
```
sinergise-anze, line 861: Nitpick, FFTR: I would suggest using a bit different naming:
```
<h4>original | no change | black & white</h4>
```
(assuming you use the suggested function for the last effect)

stories/s2l2a.stories.js
zcernigoj: Thanks, used suggested function

stories/s2l2a.stories.js
zcernigoj: Thanks, changed (a bit better than the existing).

src/mapDataManipulation/effectFunctions.ts
zcernigoj: I wrote some tests, for now just for `runColorEffectFunction` and `runCustomEffectFunction`. I plan to improve gain and gamma in the near future, so I will add them then.

README.md
zcernigoj: Changed the documentation to the simplest possible function.

src/mapDataManipulation/const.ts
```diff
@@ -9,10 +9,12 @@ export type Effects = {
...
 
-export type RgbMappingArrays = {
-  red: number[];
-  green: number[];
-  blue: number[];
+export type ImageProperties = {
+  imageData: Uint8ClampedArray;
+  imageWidth: number;
+  imageHeight: number;
+  imageFormat: string;
```
sinergise-anze, line 19: Nitpick regarding naming: 
- `image` is probably not needed as this is `ImageProperties`
- `imageData` reminds of [ImageData](https://developer.mozilla.org/en-US/docs/Web/API/ImageData), which is a different structure - maybe `rgbaArray` or just `rgba` would be a better name?

Fftr either way, not very important.

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -1,22 +1,59 @@
...
+    });
+    imgCanvas.remove();
+    return blob;
+  } catch (e) {
+    console.error(e);
+  } finally {
+    if (imgObjectUrl) {
+      window.URL.revokeObjectURL(imgObjectUrl);
+    }
+  }
```
sinergise-anze, line 56: This is not needed, maybe we should put `imgCanvas.remove()` here?

src/mapDataManipulation/mapDataManipulationUtils.ts
```diff
@@ -1,22 +1,59 @@
-import { RgbMappingArrays, Effects, ColorRange } from './const';
+import { Effects, ColorRange, ImageProperties } from './const';
 
-export function prepareRgbMappingArrays(): RgbMappingArrays {
-  return {
-    red: [...Array(256).keys()],
-    green: [...Array(256).keys()],
-    blue: [...Array(256).keys()],
-  };
+export async function getImageData(originalBlob: Blob): Promise<ImageProperties> {
```
sinergise-anze, line 3: These functions seem to be very generic. I think it would be a good idea to put them next to [existing utility functions](https://github.com/sentinel-hub/sentinelhub-js#utility-functions) to avoid duplication of efforts? Exposing them might not be needed at this time, this would serve just to avoid ending up with multiple instances of the same code in many places (please make sure that the code is not duplicated elsewhere).

src/mapDataManipulation/__tests__/effectFunctions.ts
```diff
@@ -0,0 +1,58 @@
...
+  const originalPixel = [0.2, 0.8, 1, 1];
+  const expectedPixel = [0, 1, 1, 1];
+  const effects = {
+    redRange: { from: 0.2, to: 0.8 },
+    greenRange: { from: 0.2, to: 0.8 },
+    blueRange: { from: 0.2, to: 0.8 },
+  };
+  const actualPixel = runColorEffectFunction(originalPixel, effects);
+  expect(actualPixel).toEqual(expectedPixel);
+});
```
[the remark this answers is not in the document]
zcernigoj: This test is not exactly correct as the `{ from: 0.1, to: 0.6 }` denotes that all input values <= 0.1 will transform to 0, all input values >= 0.6 will transform to 1. Input values between 0.1 and 0.6 will transform to the values between 0 an 1 (below is the equation).

```
x = input value (R or G or B component for the pixel)
oldMin = 0.1; oldMax = 0.6;
newMin = 0; newMax = 1;
let newX = newMin + ((newMax - newMin) / (oldMax - oldMin)) * (x - oldMin);
```

Changed the first test for `runColorEffectFunction` to test the border values (same as `from` and `to`). Added 2 tests. One for the values outside of the range (less than `from` and greater than `to`). And one for the values inside of the range (between `from` and `to`).

src/mapDataManipulation/const.ts
zcernigoj: Thanks, changed anyway.

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: I think this is correct (done the same way in EOB). Url needs to be revoked. I added removing of canvas to hopefully free a bit of memory.

`imgObjectUrl` is declared outside of the `try-catch` so it can be this way.
`imgCanvas` would need to be declared outside as well if we want to do it this way.

sinergise-anze: Sorry if I was not being clear - this is needed and OK in `getImageData` function, but in `getBlob` function we never assign a value to `imgObjectUrl` at all. :)

zcernigoj: Ah sorry, totally missed that. I understand now what you are saying. Thanks, will remove it here.

src/utils/canvas.ts
```diff
@@ -24,3 +24,54 @@ export async function drawBlobOnCanvas(
...
+    imgObjectUrl = window.URL.createObjectURL(originalBlob);
+    const img: any = await new Promise((resolve, reject) => {
+      const img = new Image();
+      img.onload = () => resolve(img);
+      img.onerror = reject;
+      img.src = imgObjectUrl;
+    });
+    imgCanvas.width = img.width;
+    imgCanvas.height = img.height;
+    imgCtx.drawImage(img, 0, 0);
```
zcernigoj, line 42: tried to use `drawBlobOnCanvas()` but I didn't want to change it to return img width and height (because then I would have to check if that's ok with all the other functions that use `drawBlobOnCanvas()` and it's just a bit too much out of scope for this MR I think) so I just left the duplicated code for putting the blob on canvas here.

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: Should be visible in the `/utils/canvas.ts` (moved it per suggestion below)

src/mapDataManipulation/mapDataManipulationUtils.ts
zcernigoj: I moved them to `/utils/canvas` as suggested and used `canvasToBlob` in `getBlob`. There is unfortunately some duplicated code as described below.

sinergise-anze (approved):
Nice, I think this is ready to merge! :)

PR #133, opened 2020-08-24
Choose appropriate shServiceHostname for data fusion

Ardweaden:
Data fusion only works reliably using `services.sentinel-hub` if used datasets are found at different endpoints.

Thus, if all datasets are at the same endpoint, that endpoint is used, otherwise `services.sentinel-hub` is used instead.


sinergise-anze (approved):
Nice!

PR #134, opened 2020-08-28
Fix/huge map format dimension limits

guss84:
* throw error if image format is not supported by `getHugeMap`
* Change image dimension limits according to docs

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -5,7 +5,7 @@ import { union, intersection, Geom } from 'polygon-clipping';
 
 import { BBox } from '../bbox';
 import { CRS_EPSG4326 } from '../crs';
-import { GetMapParams, ApiType, PaginatedTiles, FlyoverInterval, PreviewMode } from './const';
+import { GetMapParams, ApiType, PaginatedTiles, FlyoverInterval, PreviewMode, MimeTypes } from './const';
 import { Dataset } from './dataset';
 import { getAxiosReqParams, RequestConfiguration } from '../utils/cancelRequests';
 import { ensureTimeout } from '../utils/ensureTimeout';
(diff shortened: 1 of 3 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -89,8 +89,11 @@ export class AbstractLayer {
           'Method getHugeMap() requests that width and height of resulting image are specified',
         );
       }
+      if (params.format !== MimeTypes.JPEG || params.format !== MimeTypes.PNG) {
+        throw new Error('getHugeMap does not support' + params.format + 'only PNG and JPEG are supported');
+      }
```
sinergise-anze, line 94: - `||` should be `&&`
- spaces are missing in the error message (`...does not supportTIFFonly PNG and JPEG...`).

I would suggest:
```
      if (params.format !== MimeTypes.JPEG && params.format !== MimeTypes.PNG) {
        throw new Error('getHugeMap() only supports PNG and JPEG image formats');
      }
```

sinergise-anze: I see you changed the error message, but it is a bit misleading... `params.format` is supposed to be a MIME string, while `PNG` and `JPEG` are not... One could get such error message:
`getHugeMap does not support PNG only PNG and JPEG are supported`.

Also, we don't actually write the function name in other error messages, so we should probably remove `getHugeMap` for consistency.

The simplest solution would be to not print the `params.format`, but if you do, `PNG` and `JPEG` should be their MIME strings:
```javascript
      if (params.format !== MimeTypes.JPEG && params.format !== MimeTypes.PNG) {
        throw new Error('Format ' + params.format + ' not supported, only ' + MimeTypes.PNG + ' and ' + MimeTypes.JPEG + ' are allowed');
      }
```

guss84: Changed it to what you suggested

sinergise-anze (approved):

PR #135, opened 2020-08-28
Feature/use catalog service for S2

dgostencnik:
Use catalog service for searching for S2 tiles . At the moment both, searchIndex and catalog service are supported. If token is present, catalog service is used, otherwise searchIndex will be used.

While this works also for most of other datasets ( S1 is a bit weird and there will probably be some issues with byoc), I've decided to split it in few smaller chunks so other datasets will be added in separate PR.  

I would also appreciate any idea for less useless tests (without testing if service works correctly).


src/layer/__tests__/fixtures.findTilesS2L1C.ts
```diff
@@ -0,0 +1,325 @@
...
+    links: [
+      {
+        href: 'https://services.sentinel-hub.com/api/v1/catalog/search',
+        rel: 'self',
+        type: 'application/json',
+      },
+    ],
+    context: { limit: 50, returned: 3 },
+  };
+  /* eslint-enable */
```
sinergise-anze, line 269: This should probably be the end of the block? (not that it matters much)

dgostencnik: I wanted to disable lint just for that particular response (snake_case instead of camelCase)

src/layer/AbstractSentinelHubV1OrV2Layer.ts
```diff
@@ -135,56 +135,68 @@ export class AbstractSentinelHubV1OrV2Layer extends AbstractLayer {
...
+    toTime: Date,
+    maxCount: number | null = null,
+    offset: number | null = null,
+    reqConfig?: RequestConfiguration,
+  ): Promise<PaginatedTiles> {
+    const fetchTilesResponse = await ensureTimeout(
+      async innerReqConfig => await this.fetchTiles(bbox, fromTime, toTime, maxCount, offset, innerReqConfig),
+      reqConfig,
+    );
+    return fetchTilesResponse;
```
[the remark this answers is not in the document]
dgostencnik: Was moved and renamed as suggested

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -310,6 +311,83 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
+  ): Promise<PaginatedTiles> {
+    const response = await this.fetchTilesFromSearchIndexOrCatalog(
+      bbox,
+      fromTime,
+      toTime,
+      maxCount,
+      offset,
+      reqConfig,
+    );
+    return response;
```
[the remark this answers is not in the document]
dgostencnik: fetchTilesFromSearchIndexOrCatalog was replaced as suggested

src/layer/AbstractSentinelHubV3WithCCLayer.ts
```diff
@@ -43,36 +42,53 @@ export class AbstractSentinelHubV3WithCCLayer extends AbstractSentinelHubV3Layer
...
+    return {
+      tiles: response.data.tiles.map(tile => ({
+        geometry: tile.dataGeometry,
+        sensingTime: moment.utc(tile.sensingTime).toDate(),
+        meta: this.extractFindTilesMeta(tile),
+        links: this.getTileLinks(tile),
+      })),
+      hasMore: response.data.hasMore,
+    };
+  }
```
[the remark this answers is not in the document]
dgostencnik: Was removed

src/layer/__tests__/findTilesS2L1C.ts
```diff
@@ -0,0 +1,197 @@
...
+    const request = mockNetwork.history.post[0];
+    expect(request.data).not.toBeNull();
+    const { clipping, maxcount, maxCloudCoverage, timeFrom, timeTo } = JSON.parse(request.data);
+
+    expect(maxCloudCoverage).toEqual(expectedMaxCloudCoverage);
+    expect(maxcount).toEqual(limit);
+    expect(new Date(timeFrom)).toEqual(fromDate);
+    expect(new Date(timeTo)).toEqual(toDate);
+    expect(clipping.crs.properties.name).toEqual(CRS_EPSG4326.urn);
+    expect(clipping.crs.properties.name).toEqual(CRS_EPSG4326.urn);
```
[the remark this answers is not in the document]
dgostencnik: Duplicate was removed

src/layer/__tests__/fixtures.findTilesS2L1C.ts
sinergise-anze: I know, sorry if I wasn't clear - the block starts with `/* eslint-enable */` (correct) and ends with `/* eslint-enable */` (wrong, should be `/* eslint-disable */`).

I am assuming here that you wanted to disable linting just for this block, of course. If not, I would suggest putting `/* eslint-disable */` at the end of the file to clearly communicate with the reader that this was intentional (and then one of the `enable`s is not needed).

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -537,4 +674,11 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const shServiceHostname = this.getShServiceHostname();
     return `${shServiceHostname}api/v1/process/convertscript?datasetType=${this.dataset.shProcessingApiDatasourceAbbreviation}`;
   }
+
+  protected getFindTilesAdditionalParameters(): FindTilesAdditionalParameters {
+    return {
+      maxCloudCoverPercent: null,
+      datasetParameters: null,
+    };
+  }
```
sinergise-anze, line 683: Since this is on `AbstractSentinelHubV3Layer`, it is used by all the child layers, including S-1 on AWS, where cloud cover doesn't apply - I think S-1 shouldn't know about CC.

I think it might make sense to redefine `FindTilesAdditionalParameters` so that it can include any parameters (`Record<string, any>`) and let the child decide which ones to use. We can also define `FindTilesAdditionalParametersCC` if we want to be stricter with the keys, and then let layers decide what to use.

src/layer/S2L2ALayer.ts
```diff
@@ -37,4 +50,20 @@ export class S2L2ALayer extends AbstractSentinelHubV3WithCCLayer {
...
   }
+
+  protected extractFindTilesMetaFromCatalog(feature: Record<string, any>): Record<string, any> {
+    let result: Record<string, any> = {};
+
+    result = {
+      ...super.extractFindTilesMetaFromCatalog(feature),
+    };
+
+    if (feature.assets && feature.assets.data) {
```
sinergise-anze, line 61: With S-2 L2A you also check `&& feature.assets.data.href {`... Not sure if it is needed, just wanted to point out inconsistency if it is a bug. Fftr.

src/layer/__tests__/S2L1CLayer.ts
```diff
@@ -1,5 +1,200 @@
-import { S2L1CLayer, ApiType } from '../../index';
 import '../../../jest-setup';
+import axios from 'axios';
```
sinergise-anze, line 2: With support for catalog service on other datasets, the tests for them should also be written. However I am afraid of code duplication here - maybe we can find some smart way of using only different fixtures (responses, params,...) to test each of the datasets?

Also, IIUC, there are some issues still open on the backend which mean that this patch can't be merged yet. Maybe writing the tests would help us identify when the issues are resolved and we can merge this?

src/layer/S2L2ALayer.ts
dgostencnik: Missing check was added

src/layer/AbstractSentinelHubV3Layer.ts
dgostencnik: Type `FindTilesAdditionalParameters` was changed to `Record<string, any>`

src/layer/__tests__/findTilesS2L1C.ts
```diff
@@ -0,0 +1,197 @@
+import axios from 'axios';
```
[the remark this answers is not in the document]
dgostencnik: Tests were moved to `S2L1CLayer.ts`

package.json
```diff
@@ -69,7 +69,8 @@
     "testPathIgnorePatterns": [
       "/node_modules/",
       "/dist/",
-      "fixtures[.].*[.]ts"
+      "fixtures[.].*[.]ts",
+      "testUtils.findTiles.ts"
```
sinergise-anze, line 73: Nitpick: note that this is a regex so the correct form would be `"testUtils[.]findTiles[.]ts"`.

It might also make sense to make this generic? For example `"testUtils[.].*[.]ts"`?

Fftr either way.

stories/s1grdew.stories.js
```diff
@@ -216,8 +217,44 @@ export const getMapProcessingFromLayer = () => {
...
+      0,
+    );
+    renderTilesList(containerEl, data.tiles);
+  };
+  perform().then(() => {});
+
+  return wrapperEl;
+};
+
+export const findTilesSearchCatalog = () => {
```
sinergise-anze, line 250: I didn't investigate why, but this story issues a `searchIndex` request instead of `catalog`. The same happens with S-1 IW.

stories/s2l1c.stories.js
```diff
@@ -213,6 +216,36 @@ export const FindTiles = () => {
   return wrapperEl;
 };
 
+export const FindTilesCatalog = () => {
```
sinergise-anze, line 219: This is weird - with `FindTilesSearchIndex` the call to `searchIndex` is made (as expected), but with `FindTilesCatalog` another endpoint, `search` is used... Am I missing something?

sinergise-anze (approved):

src/layer/S2L2ALayer.ts
```diff
@@ -37,4 +50,20 @@ export class S2L2ALayer extends AbstractSentinelHubV3WithCCLayer {
         .join(''),
     };
   }
+
+  protected extractFindTilesMetaFromCatalog(feature: Record<string, any>): Record<string, any> {
+    let result: Record<string, any> = {};
+
+    result = {
+      ...super.extractFindTilesMetaFromCatalog(feature),
+    };
```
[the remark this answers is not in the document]
dgostencnik: was changed as suggested

PR #138, opened 2020-09-15
Feature/catalog service s3 s5 etc

dgostencnik:
Use catalog service with other datasets: 
- S1GRDAWSEU
- S3SLTR
- S3OLCI
- S5PL2
- Landstat8AWS
- MODIS
- BYOC

stories and tests were added for all of those

this will be merged into feature/use-catalog-service


Ardweaden (approved):
I couldn't find any problems with the code. Admittedly I haven't looked at the tests very carefully, as there really are many of them. But the other changes look ok to me.

PR #139, opened 2020-09-29
Feature/use catalog for findDatesUTC

dgostencnik:
`findDatesUTC` was using `findAvailableDates` which is part of "index" endpoint and needs to be replaced. Similar as with `findTiles` catalog is now used by `findDatesUTC` if authToken is provided. If token is not set it will still use old enpoint (that will of course eventually be replaced with an errror message).

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -607,7 +607,13 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     fromTime: Date,
     toTime: Date,
   ): Promise<Date[]> {
-    const findDatesUTCAdditionalParameters = await this.getFindDatesUTCAdditionalParameters();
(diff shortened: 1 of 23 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -487,7 +488,8 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     offset: number | null = null,
     reqConfig: RequestConfiguration,
     findTilesAdditionalParameters: FindTilesAdditionalParameters,
-  ): Promise<PaginatedTiles> {
+    distinct: string | null = null,
```
Ardweaden, line 491: Is there are reason why this is a distinct (sic) parameter and not a part of `findTilesAdditionalParameters`?

dgostencnik: No, not really. Was changed.

Ardweaden (approved):

Issue #140, opened 2020-09-30
Can this be used in a nodejs environment? I keep encountering  Reference error: window is not defined


tirth0:
(node:2528) UnhandledPromiseRejectionWarning: ReferenceError: window is not defined
    at C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:289:9
    at Generator.next (<anonymous>)
    at C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:117:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:113:12)
    at checkIfCacheApiAvailable (C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:288:12)
    at C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:308:30
    at Generator.next (<anonymous>)
    at C:\Users\tirth\Desktop\New folder (3)\node_modules\@sentinel-hub\sentinelhub-js\dist\sentinelHub.cjs.js:117:71
    at new Promise (<anonymous>)
(node:2528) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either 
by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:2528) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.     
(node:2528) UnhandledPromiseRejectionWarning: Error: Request failed with status code 400
    at createError (C:\Users\tirth\Desktop\New folder (3)\node_modules\axios\lib\core\createError.js:16:15)
    at settle (C:\Users\tirth\Desktop\New folder (3)\node_modules\axios\lib\core\settle.js:18:12)       
    at IncomingMessage.handleStreamEnd (C:\Users\tirth\Desktop\New folder (3)\node_modules\axios\lib\adapters\http.js:202:11)
    at IncomingMessage.emit (events.js:215:7)
[...]

PR #141, opened 2020-09-30
Check if window object exists and don't use window.URL....

zcernigoj:
`getImageProperties` used `window.URL.createObjectURL` and `window.URL.revokeObjectURL` 

`checkIfCacheApiAvailable` didn't check if `window` object exists

Both caused which problems when using the library in `node.`


sinergise-anze (approved):

sinergise-anze:
These changes are OK and they solve #140, as far as I can tell (the bug no longer occurs when running `example/node/index.js`).

PR #142, opened 2020-10-01
Set evalscript in updateLayerFromServiceIfNeeded

Ardweaden:
This is a necessary change for the evalscript parsing to work in EOB3, but I think there is no harm adding it here.

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -671,6 +671,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     await ensureTimeout(async innerReqConfig => {
       const layerParams = await this.fetchLayerParamsFromSHServiceV3(innerReqConfig);
       this.legend = layerParams['legend'] ? layerParams['legend'] : null;
+      this.evalscript = layerParams['evalscript'] ? layerParams['evalscript'] : null;
       // this is a hotfix for `supportsApiType()` not having enough information - should be fixed properly later:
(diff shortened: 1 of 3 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -671,6 +671,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     await ensureTimeout(async innerReqConfig => {
       const layerParams = await this.fetchLayerParamsFromSHServiceV3(innerReqConfig);
       this.legend = layerParams['legend'] ? layerParams['legend'] : null;
+      this.evalscript = layerParams['evalscript'] ? layerParams['evalscript'] : null;
```
sinergise-anze, line 674: I would suggest moving this line two lines below (right below the `dataProduct` line) so that the comment about hotfix applies to this line too (it is the same problem). Not the best solution, I agree... :-/

Also, note that this change does *not* get used in subclasses `S1GRDAWSEULayer` and `BYOCLayer` because of the way code is organized, so we probably want to put it there too?

Ardweaden: I don't think it's the *same* problem... It's distantly related I'd say. I intentionally put it above the comment. Are you sure we should put it there?

Thanks for noticing, added!

sinergise-anze: True, it is related, not quite the same... Then maybe a better idea would be adding a short comment explaining why this is here? Still, not very important, fftr.

sinergise-anze (approved):

PR #143, opened 2020-10-06
Fix/catalog search limit

dgostencnik:
Catalog service uses `limit` param to limit number of results. This value has been recently significantly decreased   (from 10000 to 100). As a result it is no longer possible to get all results in single request, so iterating over pages was added for findDatesUTC.


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -515,6 +520,11 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       collections: [catalogCollectionId],
       limit: maxCount,
     };
+
+    if (maxCount === null) {
+      delete payload.limit;
+    }
```
zcernigoj, line 526: I would flip this `if`, so it's more consistent with others below.

zcernigoj (approved):

src/layer/AbstractSentinelHubV3Layer.ts
dgostencnik: Was changed as suggested

PR #147, opened 2020-12-09
Feature/add DEM datasets

dgostencnik:
According to documentation https://docs.sentinel-hub.com/api/latest/data/dem/  DEM is available on 2 locations so 2 datasets were created (not sure about those names, but I was not able to come up with something better)
- DATASET_AWS_DEM for `services.sentinel-hub.com/api` enpoint
- DATASET_AWSUS_DEM for `services-uswest2.sentinel-hub.com/api`

Akternative and perhaps better opition would be to have 4 datasets
- MAPZEN_US (us endpoint)
- MAPZEN_EU 
- COPERNICUS_30
- COPERNICUS_90

As search  is not supported neither by index nor by catalog service,  `findTiles` and `findDatesUTC` are mocked and return one tile per day with bbox as geometry.

DEM has also 2 additional params, `egm` and `clampNegative`, but they are not handled yet.


src/index.ts
```diff
@@ -8,6 +8,7 @@ import {
   PreviewMode,
   MosaickingOrder,
   Interpolator,
+  DEMInstanceType,
```
sinergise-anze, line 11: Maybe `DEMType` would be a better variable name? At least I associate "instance" with WMS... Fftr.

src/layer/dataset.ts
```diff
@@ -209,6 +209,20 @@ export const DATASET_AWS_DEM: Dataset = {
...
   datasetParametersType: null,
+  shServiceHostname: 'https://services.sentinel-hub.com/',
+  searchIndexUrl: null,
+  findDatesUTCUrl: null,
+  orbitTimeMinutes: null,
+  minDate: null,
+  maxDate: null,
+};
+
+export const DATASET_AWSUS_DEM: Dataset = {
```
sinergise-anze, line 220: The suggested naming (Mapzen / Copernicus) makes much more sense IMHO, but I'm not sure how much additional work this entails? 

Assigning back the PR in case you decide to change the naming (didn't dive deeply yet), if you decide against it, please assign back.

src/layer/dataset.ts
```diff
@@ -209,6 +209,20 @@ export const DATASET_AWS_DEM: Dataset = {
   shWmsEvalsource: 'DEM',
   shProcessingApiDatasourceAbbreviation: 'DEM',
   datasetParametersType: null,
+  shServiceHostname: 'https://services.sentinel-hub.com/',
+  searchIndexUrl: null,
+  findDatesUTCUrl: null,
+  orbitTimeMinutes: null,
+  minDate: null,
+  maxDate: null,
+};
```
sinergise-anze, line 218: Changing the existing dataset definitions is a backwards incompatible change. It is very likely that nobody uses this yet, but maybe it is better to avoid taking chances... I would suggest changing names so that this one (`DATASET_AWS_DEM`) is not used, and then just linking `DATASET_AWS_DEM` to the appropriate new name. Not a strong opinion though, as this probably wasn't used by anyone yet. Fftr.

src/layer/dataset.ts
dgostencnik: As I said there are 2 reasons I don't like idea of having 4 datasets
-  makeLayers won't work correctly without adding `demInstance` to json capabilites. Based on previous experience it is probably safe to say that this will not happen or at least not any time soon (discussions about adding collectionId, s5 productType, ...)
- I don't see a point of doing it. To me this DEM  (DEM dataset + demInstance combo)  is same use case as S5 datasets with  productType and afaik we have one S5 dataset not 15 of them. Don't get me wrong, it is small change if you ignore issue mentioned in first bullet.

src/index.ts
sinergise-anze: I see now that this is actually in line with the naming on SH service, resolving.

src/layer/AbstractDEMLayer.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    const intervalStart = fromTime ? moment.utc(fromTime) : moment.utc(toTime);
+    const dateIterator = moment.utc(toTime);
+
+    while (dateIterator.isAfter(intervalStart)) {
+      if (maxCount && tiles.length >= maxCount) {
+        hasMore = true;
+        break;
+      }
+      tiles.push({ geometry: this.bboxToPolygon(bbox), sensingTime: dateIterator.toDate(), meta: {} });
```
sinergise-anze, line 101: Since this is not so obvious, maybe a comment (above the function signature?) explaining the hack with geometry could be explained?

src/layer/AbstractDEMLayer.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    return { tiles: tiles, hasMore: hasMore };
+  }
+
+  public async findDatesUTC(
+    bbox: BBox,
+    fromTime: Date,
+    toTime: Date,
+    reqConfig?: RequestConfiguration, // eslint-disable-line @typescript-eslint/no-unused-vars
```
sinergise-anze, line 112: I think this parameter should be used, since we are making network requests?

src/layer/dataset.ts
sinergise-anze: I see your point. It seems that the 3 DEM products at services are treated as a single dataset, so let's follow that. Thank you for the explanation, resolving.

sinergise-anze (commented):
Apart from not using `reqConfig` in `findDatesUTC`, it looks great!

src/layer/AbstractDEMLayer.ts
sinergise-anze: Nevermind, long day. :) Resolving.

src/layer/AbstractDEMLayer.ts
```diff
@@ -0,0 +1,118 @@
...
+  ): Promise<PaginatedTiles> {
+    const tiles = [];
+    let hasMore = false;
+
+    tiles.push({
+      geometry: this.bboxToPolygon(bbox),
+      sensingTime: moment
+        .utc(toTime)
+        .startOf('day')
```
sinergise-anze, line 101: Maybe it would be better to remove this line? Start of the day might be outside of fromTime/toTime interval.

sinergise-anze (approved):

PR #150, opened 2020-12-15
2 Additional DEM layer properties

dgostencnik:
DEM datasources have 2 additional properties: `egm` and `clampNegative`. They should't be ignored as they can be set in dashboard and can affect visualization (at least `clampNegative`).

### DIFF
--- a/src/layer/AbstractDEMLayer.ts
+++ b/src/layer/AbstractDEMLayer.ts
@@ -48,10 +48,10 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
         if (!this.demInstance) {
           this.demInstance = layerParams['demInstance'] ? layerParams['demInstance'] : DEMInstanceType.MAPZEN;
         }
-        if (isBooleanNull(this.clampNegative)) {
-          this.clampNegative = layerParams['clampNegative'] ? layerParams['clampNegative'] : false;
+        if (!isDefined(this.clampNegative)) {
+          this.clampNegative = layerParams['clampNegative'] ? layerParams['clampNegative'] : null;
         }
-        if (isBooleanNull(this.egm)) {
+        if (!isDefined(this.egm)) {
           //this in not a typo. Configuration service returns `EGM`, process api accepts `egm`
           this.egm = layerParams['EGM'] ? layerParams['EGM'] : false;
         }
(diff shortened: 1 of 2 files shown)


src/layer/AbstractDEMLayer.ts
```diff
@@ -116,3 +146,5 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
     return tiles.map(tile => moment.utc(tile.sensingTime).toDate());
   }
 }
+
+const isBooleanNull = (value: boolean): boolean => value === null || value == undefined;
```
Ardweaden, line 150: I think `value` is not  a `boolean`, it's of types `null` or `undefined`.  I would assume `isDefined` or smth would be a more appropriate name.

src/layer/AbstractDEMLayer.ts
```diff
@@ -25,21 +25,36 @@ export interface ConstructorParameters {
...
-
-      if (!this.demInstance) {
+      //update properties defined on parent layer
+      await super.updateLayerFromServiceIfNeeded(innerReqConfig);
+      //update DEM specific properties if they're not set
+      if (!this.demInstance || isBooleanNull(this.egm) || isBooleanNull(this.clampNegative)) {
         const layerParams = await this.fetchLayerParamsFromSHServiceV3(innerReqConfig);
-        this.demInstance = layerParams['demInstance'] ? layerParams['demInstance'] : DEMInstanceType.MAPZEN;
+        if (!this.demInstance) {
+          this.demInstance = layerParams['demInstance'] ? layerParams['demInstance'] : DEMInstanceType.MAPZEN;
```
Ardweaden, line 49: I don't think it's a good idea to use default value. This could give false impression that the layer is defined in a certain way on the service. IMO all of them should be just set to `null` (also `clampNegative` and `egm`).

src/layer/AbstractDEMLayer.ts
dgostencnik: Function was renamed to `isDefined`

src/layer/AbstractDEMLayer.ts
dgostencnik: Defaults were removed and check was added to ensure `fetchLayerParamsFromSHServiceV3` is not called everytime because of null values.

Ardweaden (approved):

PR #152, opened 2020-12-19
Allow replacing hostnames

sinergise-anze:
This PR allows apps to redirect traffic to different domains, with the purpose of using staging/testing SH environments.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
 CLIENT_ID=
 CLIENT_SECRET=
 
+ALTERNATIVE_HOSTNAME=
+
 # Env var INSTANCE_ID should be set to the ID of some instance in Sentinel Hub
 # Configurator, which contains all the layers listed below it.
 INSTANCE_ID=
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -22,11 +22,14 @@ declare module 'axios' {
   }
 }
 
-export const registerAxiosCacheRetryInterceptors = (): any => {
+export const registerInitialAxiosInterceptors = (): any => {
   findAndDeleteExpiredCachedItems();
+  // note that the interceptors are called in reverse order in which they are registered - last
+  // defined interceptor is called first:
   axios.interceptors.request.use(logCurl, error => Promise.reject(error));
   axios.interceptors.request.use(fetchCachedResponse, error => Promise.reject(error));
   axios.interceptors.response.use(saveCacheResponse, error => retryRequests(error));
+  // additionally, `registerHostnameReplacing()` registers another request interceptor (if/when called)
 };
 
 const logCurl = async (config: any): Promise<any> => {
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -24,12 +24,12 @@ declare module 'axios' {
 
 export const registerInitialAxiosInterceptors = (): any => {
   findAndDeleteExpiredCachedItems();
-  // note that the interceptors are called in reverse order in which they are registered - last
-  // defined interceptor is called first:
+  // - the interceptors are called in reverse order in which they are registered - last
+  //   defined interceptor is called first
+  // - some interceptors might also be added in other places (`registerHostnameReplacing()`)
   axios.interceptors.request.use(logCurl, error => Promise.reject(error));
   axios.interceptors.request.use(fetchCachedResponse, error => Promise.reject(error));
   axios.interceptors.response.use(saveCacheResponse, error => retryRequests(error));
-  // additionally, `registerHostnameReplacing()` registers another request interceptor (if/when called)
 };
 
 const logCurl = async (config: any): Promise<any> => {
--- a/stories/s2l2a.stories.js
+++ b/stories/s2l2a.stories.js
@@ -9,6 +9,7 @@ import {
   ApiType,
   CRS_EPSG3857,
   registerHostnameReplacing,
+  requestAuthToken,
 } from '../dist/sentinelHub.esm';
 
 if (!process.env.INSTANCE_ID) {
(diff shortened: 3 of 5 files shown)


.env.example
```diff
@@ -1,6 +1,8 @@
 CLIENT_ID=
 CLIENT_SECRET=
 
+STAGE_HOSTNAME=
```
zcernigoj, line 4: Would  `ALTERNATIVE_HOSTNAME` be a better name or is it ok that it's so specific?

src/utils/axiosInterceptors.ts
```diff
@@ -22,11 +22,14 @@ declare module 'axios' {
...
 
-export const registerAxiosCacheRetryInterceptors = (): any => {
+export const registerInitialAxiosInterceptors = (): any => {
   findAndDeleteExpiredCachedItems();
+  // note that the interceptors are called in reverse order in which they are registered - last
+  // defined interceptor is called first:
   axios.interceptors.request.use(logCurl, error => Promise.reject(error));
   axios.interceptors.request.use(fetchCachedResponse, error => Promise.reject(error));
   axios.interceptors.response.use(saveCacheResponse, error => retryRequests(error));
+  // additionally, `registerHostnameReplacing()` registers another request interceptor (if/when called)
```
zcernigoj, line 32: Is this code here a leftover or is some code missing here?

If it's meant to be here, I think it would need to be changed because now it suggests there should be some code after it.

zcernigoj:
If I set the hostname without `https://` in the `.env` file, I get the error below for `Get Map Processing` story. If I set the hostname with `https://`, that story works.

```
Access to XMLHttpRequest at 'https://<hostname>/api/v1/process' from origin 'http://localhost:6006' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: It does not have HTTP ok status
```

In both cases I get `s2l2a.stories.js:361 Uncaught (in promise) ReferenceError: requestAuthToken is not defined` for the `Get Map Processing Staging` story.

stories/s2l2a.stories.js
```diff
@@ -341,6 +342,55 @@ export const GetMapProcessingEvalscripturlVersion2 = () => {
   return wrapperEl;
 };
 
+export const GetMapProcessingStaging = () => {
+  if (!process.env.CLIENT_ID || !process.env.CLIENT_SECRET) {
```
zcernigoj, line 346: I'm not sure the same client id and secret work for both stage and production. Did you test?

Just want to make sure. FFTR.

.env.example
sinergise-anze: Thank you for the suggestion, yes, much better name! Changed.

src/utils/axiosInterceptors.ts
sinergise-anze: It was meant to be here, but apparently it wasn't as clear as I hoped it would be. :)

I have now changed the comments, is this better?

sinergise-anze:
> **Things related to stories**
> 
>     * If I set the `hostname` without `https://` in the `.env` file, I get the error below for `Get Map Processing` story. If I set the hostname with `https://`, that story works.
>       ```
>       Access to XMLHttpRequest at 'https://<hostname>/api/v1/process' from origin 'http://localhost:6006' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: It does not have HTTP ok status
>       ```
> 
>     * In both cases (`hostname` or `hostname` with added `https://`) I get `s2l2a.stories.js:361 Uncaught (in promise) ReferenceError: requestAuthToken is not defined` for the newly added `Get Map Processing Staging` story.

It seems I have made some last-minute changes just before commiting that broke the build, sorry about that.

The hostname should always be just the domain (without the `https://` part) - it should work now.

stories/s2l2a.stories.js
sinergise-anze: If I understand correctly it is possible to fake whatever credentials in staging, though I don't know for sure. But we can always change if this is not enough.

.env.example
zcernigoj: Nice. Resolving

stories/s2l2a.stories.js
zcernigoj: Ok, I agree. Resolving

src/utils/axiosInterceptors.ts
zcernigoj: Thanks, it is better now :) :+1: 

Resolving

stories/s2l2a.stories.js
```diff
@@ -341,6 +343,59 @@ export const GetMapProcessingEvalscripturlVersion2 = () => {
...
+        width: 512,
+        height: 512,
+        format: MimeTypes.JPEG,
+      };
+      const imageBlob = await layerS2L2A.getMap(getMapParams, ApiType.PROCESSING);
+      img.src = URL.createObjectURL(imageBlob);
+    } finally {
+      // reset auth token back, so that we don't break other stories:
+      registerHostnameReplacing('services.sentinel-hub.com', 'services.sentinel-hub.com');
+      const authTokenServices = await requestAuthToken(clientId, clientSecret);
```
zcernigoj, line 390: `clientId` and `clientSecret` are not defined. I think `process.env.CLIENT_ID` and `process.env.CLIENT_SECRET`  can be used instead.

stories/s2l2a.stories.js
```diff
@@ -341,6 +343,59 @@ export const GetMapProcessingEvalscripturlVersion2 = () => {
...
+      setAuthToken(authTokenServices);
+      printOut('Auth token reset.');
```
zcernigoj, line 392: `printOut()` is only available in node example (`example/node/index.js`). I think changing this to `console.log()` is the quickest solution (no need to complicate).

sinergise-anze: Thanks for spotting this, fixed.

stories/s2l2a.stories.js
zcernigoj: Was fixed. Resolving

stories/s2l2a.stories.js
zcernigoj: Was fixed. Resolving

zcernigoj (approved):
Ready to merge :+1:

PR #153, opened 2021-01-20
Expose getAuthToken

Ardweaden:


src/index.ts
```diff
@@ -111,6 +111,7 @@ export {
   setAuthToken,
   isAuthTokenSet,
   requestAuthToken,
+  getAuthToken,
```
sinergise-anze, line 114: Initial decision was to leave out `getAuthToken` - app should keep its authentication token(s) anyway, because it needs to refresh them when needed, so there should be no need for it. The lack of this function also makes it very clear that "source of truth" for app tokens is the app state, not the library state.

I am not completely opposed to this change, but can you maybe provide some feedback why it is needed?

Ardweaden: It is not, I am closing this MR.

Ardweaden:
This change is not necessary

PR #154, opened 2021-01-27
Check if canvas exceeds the size limit 

Ardweaden:
If the canvas exceed the size limit for the browser, `canvas.toBlob` returns `null`. 

Run `canvas.toBlob` before fetching images, to test if it produces a blob.

https://www.npmjs.com/package/canvas-size#test-results

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -114,7 +114,7 @@ export class AbstractLayer {
       const isCanvasValid = await validateCanvasDimensions(canvas);
       if (!isCanvasValid) {
         throw new Error(
-          `Image dimensions (${width}x${height}) exceeds the canvas size limit for this browser.`,
+          `Image dimensions (${width}x${height}) exceed the canvas size limit for this browser.`,
         );
       }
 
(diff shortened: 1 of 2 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -111,6 +111,13 @@ export class AbstractLayer {
       canvas.height = height;
       const ctx = canvas.getContext('2d');
 
+      const isCanvasValid = await validateCanvasDimensions(canvas);
+      if (!isCanvasValid) {
+        throw new Error(
+          `Image dimensions (${width}x${height}) exceeds the canvas size limit for this browser.`,
```
sinergise-anze, line 117: Typo: `exceeds` -> `exceed`

Feel free to resolve.

src/utils/canvas.ts
```diff
@@ -75,3 +75,11 @@ export async function getBlob(imageProperties: ImageProperties): Promise<Blob> {
     imgCanvas.remove();
   }
 }
+
+export async function validateCanvasDimensions(canvas: HTMLCanvasElement): Promise<boolean> {
```
sinergise-anze, line 79: Looks great!

Maybe adding a comment that would explain why this check is needed would help future readers?

For example:
```javascript
// If the canvas exceeds the size limit for the browser, canvas.toBlob returns null.
// This function tests for this condition so that we can warn user.
```

Minor, fftr.

sinergise-anze (approved):

src/utils/canvas.ts
Ardweaden: I added just the first part, that's sufficient imo

src/layer/AbstractLayer.ts
Ardweaden: Thanks, fixed.

PR #155, opened 2021-02-01
Add/orthorectification parameter

dthiex:
Adds demInstance to S1 parameters so the set DEM is used for orthorectification and radiometric terrain correction.


src/layer/const.ts
```diff
@@ -226,3 +227,10 @@ export enum DEMInstanceType {
   COPERNICUS_30 = 'COPERNICUS_30',
   COPERNICUS_90 = 'COPERNICUS_90',
 }
+
+export enum DEMInstanceTypeOrthorectification {
```
dgostencnik, line 231: This looks as duplication but I assume there is some good reason for this. fftr

src/layer/S1GRDAWSEULayer.ts
```diff
@@ -139,6 +145,7 @@ export class S1GRDAWSEULayer extends AbstractSentinelHubV3Layer {
     }
     payload.input.data[0].processing.backCoeff = this.backscatterCoeff;
     payload.input.data[0].processing.orthorectify = this.orthorectify;
+    payload.input.data[0].processing.demInstance = this.demInstanceType;
```
dgostencnik, line 148: I'm not sure if it makes sense to always add  `demInstance` to payload. It might be better to set it only when ortorectification is enabled. fftr

dgostencnik (approved):

src/layer/S1GRDAWSEULayer.ts
dthiex: Probably as demInstanceType requires orthorectify to be set true. I updated it, can you please again.

src/layer/const.ts
dthiex: The reason why this was added as separate variable is because for the orthorectification the copernicus dem with 10 meter resolution can be used but it's not available as "normal" dem instance.

PR #157, opened 2021-02-04
Fix/parallel requests caching

sinergise-anze:
This MR solves a problem when multiple requests to the same resource, issued in parallel at (almost) the same time, didn't use caching mechanism. The problem was that by the time the first response came, the other requests have already started.

The easiest solution was to delay requests if there is an ongoing request with the same cache key.

Apart from that:
- `generateCacheKey` does not need to be `async`

### DIFF
--- a/src/utils/cacheHandlers.ts
+++ b/src/utils/cacheHandlers.ts
@@ -11,38 +11,61 @@ export type CacheConfig = {
   targets?: CacheTargets;
 };
 
+// Even though we have caching enabled, if we fire 10 (equal) requests in parallel, they will
+// still all be executed - because by the time first response is saved in cache, the other 9
+// requests are already made too. To combat this, we save cacheKeys of ongoing requests and
+// simply delay new requests with the same cacheKey.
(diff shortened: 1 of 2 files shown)


src/utils/cacheHandlers.ts
```diff
@@ -54,25 +77,33 @@ export const saveCacheResponse = async (response: AxiosResponse): Promise<any> =
...
+    // before saving response, set an artificial header that tells when it should expire:
+    const expiresMs = new Date().getTime() + response.config.cache.expiresIn * 1000;
+    response.headers = {
+      ...response.headers,
+      [EXPIRY_HEADER_KEY]: expiresMs,
+    };
+    shCache.set(response.config.cacheKey, response);
     return response;
+  } finally {
+    // if response.confic.cacheKey was there, we *must* remove it from the list
```
guss84, line 102: typo in the comment

sinergise-anze: Thanks, fixed! :grin:

guss84 (approved):

PR #158, opened 2021-02-04
Feature/default reqs config

sinergise-anze:
This MR introduces `setDefaultRequestsConfig()` function. It allows setting the settings app-wide, so that they don't need to be passed through `reqConfig` parameter every time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,15 @@ A list of targets can be provided which is ordered by priority, and the first av
     },
   };
 ```
+
+If a default requests configuration object is specified, it will be used for any key which is not set explicitly:
+
+```javascript
+  setDefaultRequestsConfig({
+    retries: 2,
+  });
+```
+
 ## Getting basic statistics and histogram
 
 Getting basic statistics (mean, min, max, standard deviation) and a histogram for a geometry (Polygon or MultiPolygon).
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ Services limit the size of the output image per request (2500px in each directio
 IMPORTANT: be careful with the image sizes as a big image could consume a lot of processing units. There is no limit imposed by this method.
 
 ```javascript
-  const imageBlob = await layer.getHugeMap(getMapParams, ApiType.PROCESSING, reqConfig);
+  const imageBlob = await layer.getHugeMap(getMapParams, ApiType.PROCESSING, requestsConfig);
 ```
 
 ## Searching for data
@@ -306,9 +306,9 @@ try {
 }
 ```
 
-To enable caching for the requests, one can add `expiresIn` to the requests configuration object. The values are in seconds.
-```
-// cache is valid for 30 minutes
+To enable caching for the requests, one can add `expiresIn` to the requests configuration object. The values are in seconds. Value `0` disables caching.
+```javascript
+// cache is valid for 30 minutes:
 const requestsConfig = {
   cache: {
     expiresIn: 1800,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -306,52 +306,54 @@ try {
...
   const requestsConfig = {
     cache: {
       expiresIn: 5000,
       targets: [CacheTarget.CACHE_API],
     },
   };
 ```
-```
+
+```javascript
```
zcernigoj, line 329: Do these two examples look better when separate or can they be in the same code block?

src/utils/cancelRequests.ts
```diff
@@ -39,19 +41,24 @@ export const getAxiosReqParams = (
...
 
-  if (reqConfig.cancelToken) {
-    axiosReqConfig.cancelToken = reqConfig.cancelToken.getToken();
+  if (reqConfigWithDefault.cancelToken) {
+    axiosReqConfig.cancelToken = reqConfigWithDefault.cancelToken.getToken();
   }
-  axiosReqConfig.retries = reqConfig.retries;
-  if (reqConfig.cache) {
-    axiosReqConfig.cache = reqConfig.cache;
+  axiosReqConfig.retries = reqConfigWithDefault.retries;
```
dgostencnik, line 56: I think you could also check if property `retries` is set/defined on `reqConfigWithDefault` object to avoid setting it as undefined  (same a as above),

src/utils/cancelRequests.ts
```diff
@@ -39,19 +41,24 @@ export const getAxiosReqParams = (
     cache: defaultCache,
   };
 
-  if (!reqConfig) {
+  const reqConfigWithDefault = {
+    ...getDefaultRequestsConfig(),
+    ...reqConfig,
+  };
+
+  if (!reqConfigWithDefault) {
```
dgostencnik, line 49: `reqConfigWithDefault` will never be null or undefined so it should also probably check if object has at least one key. Not that it matters.

src/utils/defaultReqsConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { RequestConfiguration } from './cancelRequests';
```
dgostencnik, line 1: `RequestConfiguration` and  `getAxiosReqParams`  should probably be moved from cancelRequests to some other location at some point as they are no longer used only for canceling requests .fftr

README.md
sinergise-anze: I agree it did look a bit weird, so I added another sentence in between. Thanks, resolving.

src/utils/cancelRequests.ts
sinergise-anze: True, thank you, fixed! 

(I haven't really changed this code, only renamed the variable - but it makes sense to fix this)

src/utils/cancelRequests.ts
sinergise-anze: ~~I agree it doesn't... but anyway, fixed the condition.~~ Since this condition never triggered before, and since it doesn't add anything, I have simply removed it now. Thank you for bringing it to my attention! Fixed.

README.md
zcernigoj: Thanks, looks good to me

zcernigoj (commented):
It looks ok to me with the changes made on @dgostencnik commented.

src/utils/defaultReqsConfig.ts
sinergise-anze: I agree - but I'm reluctant to do this change in this MR. I wanted the change to be small (because it was a hotfix), and now that it is done and tested, I'd prefer if the refactoring was done sometime in the future.

PR #159, opened 2021-02-11
Fix: parsing GetCapabilities, ignore layers without Name

sinergise-anze:
What started out as a simple small MR, turned out to be a bigger issue:
- with recent changes to GIBS, `GetCapabilites` response includes layers without `<Name>`; this patch makes `makeLayers` ignore such layers, as we can't get their ID
- we also add support for nested layers (otherwise the list of layers from GIBS is empty)
- we use the same function for getting layers from `GetCapabilities` in all places (code deduplication) to make these changes effective everywhere

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -245,20 +245,18 @@ export class LayersFactory {
     reqConfig: RequestConfiguration,
   ): Promise<AbstractLayer[]> {
     const parsedLayers = await fetchLayersFromGetCapabilitiesXml(baseUrl, reqConfig);
-    const layersInfos = parsedLayers
-      .filter(layerInfo => layerInfo.Name)
-      .map(layerInfo => ({
-        layerId: layerInfo.Name[0],
-        title: layerInfo.Title[0],
-        description: layerInfo.Abstract ? layerInfo.Abstract[0] : null,
-        dataset: null,
-        legendUrl:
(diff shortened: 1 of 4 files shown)


src/layer/LayersFactory.ts
```diff
@@ -244,19 +244,21 @@ export class LayersFactory {
...
-      legendUrl:
-        layerInfo.Style && layerInfo.Style[0].LegendURL
-          ? layerInfo.Style[0].LegendURL[0].OnlineResource[0]['$']['xlink:href']
-          : layerInfo.Layer && layerInfo.Layer[0].Style && layerInfo.Layer[0].Style[0].LegendURL
-          ? layerInfo.Layer[0].Style[0].LegendURL[0].OnlineResource[0]['$']['xlink:href']
-          : null,
-    }));
+    const parsedLayers = await fetchLayersFromGetCapabilitiesXml(baseUrl, reqConfig);
+    const layersInfos = parsedLayers
+      .filter(layerInfo => layerInfo.Name)
```
dgostencnik, line 249: layers without name are already removed in `fetchLayersFromGetCapabilitiesXml`.fftr

dgostencnik (approved):

src/layer/LayersFactory.ts
sinergise-anze: Thank you, fixed!

PR #160, opened 2021-02-25
Support S-1 orthorectify and demInstanceType with data fusion

sinergise-anze:
Only a select few layer parameters are passed to Process API when doing the data fusion. This should be fixed systematically, but for now we need the S-1 orthorectification parameters to be passed, so this is the minimal (and safest) change that achieves that.

### DIFF
--- a/src/layer/ProcessingDataFusionLayer.ts
+++ b/src/layer/ProcessingDataFusionLayer.ts
@@ -95,12 +95,22 @@ export class ProcessingDataFusionLayer extends AbstractSentinelHubV3Layer {
           datasource.dataFilter.mosaickingOrder = layerInfo.layer.mosaickingOrder;
         }
 
+        // note that we should be using updateProcessingGetMapPayload or sth. similar here, this is just a
+        // temporary band-aid which lets us quickly use datafusion:
         if (layerInfo.layer.upsampling) {
           datasource.processing.upsampling = layerInfo.layer.upsampling;
         }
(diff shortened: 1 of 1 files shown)


src/layer/ProcessingDataFusionLayer.ts
```diff
@@ -95,13 +95,23 @@ export class ProcessingDataFusionLayer extends AbstractSentinelHubV3Layer {
           datasource.dataFilter.mosaickingOrder = layerInfo.layer.mosaickingOrder;
         }
 
+        // note that we should be using updateProcessingGetMapPayload or sth. similar here, this is just a
```
Ardweaden, line 98: For some reason I can't add comments to the unchanged parts of the code.

However, would it work to just run `updateProcessingGetMapPayload` of the `bogusFirstLayer` on the `payload` in line 66 to have orthorectify  on the "primary layer"?

sinergise-anze: I have tried using `updateProcessingGetMapPayload` in a separate branch, but there are two problems with it:
- it is `protected` (and I'm reluctant to change it to `public`)
- it assumes that there is only a single dataset

In the end the branch touched too many places and I wasn't satisfied with how it looked, so I scraped it. I do agree with you though and would like to rearrange the code again now that there is a bit more time for it.

Ardweaden (approved):

PR #161, opened 2021-02-25
Fix/datafusion layer params3

sinergise-anze:
This PR is similar to #160, but tries to solve setting data fusion Process API parameters more properly (that is, layers themselves are now responsible for setting the params).

- to do that, I needed to make `updateProcessingGetMapPayload` public - however, I have prepended it with `_` to discourage its use outside of the library
- `evalscriptUrl` parameter in `ProcessingDataFusionLayer` constructor was mandatory; I have added `?` to avoid having to declare it in tests
- I didn't dare remove `upsampling` and `downsampling` from `ProcessingDataFusionLayer.getMap()` - they are not covered with tests and I am not certain that I can remove them
- I have rearranged the existing test for `ProcessingDataFusionLayer` a bit to allow adding a new one in a clearer manner

### DIFF
--- a/src/layer/AbstractDEMLayer.ts
+++ b/src/layer/AbstractDEMLayer.ts
@@ -80,24 +80,27 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
     return getMapValue;
   }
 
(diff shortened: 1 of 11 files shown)


zcernigoj (approved):
Can't find any problem, so I guess this can be merged.

PR #162, opened 2021-03-01
feature / add output responses support

zcernigoj:
Adds support for setting the output responses in `getMapParams`.

Currently limited to only 1, because with setting multiple, the response from SH is a tar, and some parts of sentinelhub-js don't know how to deal with the tar yet.

Also, this feature is not supported on WMS, so an exception is thrown.

Added tests and text to readme.

[docs for output response object in evalscript](https://docs.sentinel-hub.com/api/latest/evalscript/v3/#output-object-properties)
[example 1](https://docs.sentinel-hub.com/api/latest/evalscript/v3/#examples-1)
[example 2](https://docs.sentinel-hub.com/api/latest/data/sentinel-2-l1c/examples/#ndvi-image--and--value-multi-part-response-png-and-geotiff)


README.md
```diff
@@ -160,6 +160,23 @@ It is also possible to determine whether a layer supports a specific ApiType:
...
+
+```javascript
+  const getMapParams = {
+    bbox: bbox,
+    fromTime: new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0)),
+    toTime: new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59)),
+    width: 512,
+    height: 512,
+    format: MimeTypes.JPEG,
+    outputResponses: [{ id: 'default', format: MimeTypes.JPEG }],
```
grafolean, line 174: Since one can only request a single output at a time (well, barring `tar`, but...), wouldn't it make  more sense to simply add `outputResponseId` next to `format`?

```javascript
  const getMapParams = {
    bbox: bbox,
    fromTime: new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0)),
    toTime: new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59)),
    width: 512,
    height: 512,
    format: MimeTypes.JPEG,
    outputResponseId: 'default',
};
```

sinergise-anze, line 174: Since one can only request a single output at a time (well, barring `tar`, but...), wouldn't it make  more sense to simply add `outputResponseId` next to `format`?

```javascript
const getMapParams = {
    bbox: bbox,
    fromTime: new Date(Date.UTC(2018, 11 - 1, 22, 0, 0, 0)),
    toTime: new Date(Date.UTC(2018, 12 - 1, 22, 23, 59, 59)),
    width: 512,
    height: 512,
    format: MimeTypes.JPEG,
    outputResponseId: 'default',
};
```

zcernigoj: Thanks. Talked live and came up with the change that I implemented. Also added that an exception is thrown if a user wants to use this with WMS.

src/layer/processing.ts
```diff
@@ -98,6 +98,13 @@ export function createProcessingPayload(
 ): ProcessingPayload {
   const { bbox } = params;
 
+  // Processing API supports setting multiple output responses, which are then returned as TAR archive.
+  // Sentinelhub-js can't deal with manipulating files inside the tar yet,
+  // so we only allow setting one output response id.
+  if (params.outputResponseId === '') {
+    throw new Error('outputResponseId most not be empty');
+  }
+
```
sinergise-anze, line 107: Unless I'm missing something, the comment doesn't match the code? 

Also, the code handles a very specific case where `params.outputResponseId === ''`, which might not need to be handled (it can be set to `"default"` lower). It is a very specific corner case which will probably never arise, and if it does, one could argue that  using `"default"` is what the user would have wanted to do anyway. If you agree, I would suggest removing these lined altogether (though the comment might be useful if put elsewhere - `src/layer/const.ts` line 35 maybe?).

Very minor though, fftr either way.

src/layer/processing.ts
```diff
@@ -123,14 +130,19 @@ export function createProcessingPayload(
...
+              identifier: params.outputResponseId,
+              format: { type: params.format },
+            },
+          ]
+        : [
+            {
+              identifier: 'default',
+              format: { type: params.format },
+            },
+          ],
```
sinergise-anze, line 145: Nitpick: I think this would be more readable:
```javascript
      [
        {
          identifier: params.outputResponseId ? params.outputResponseId : 'default',
          format: { type: params.format },
        },
      ]
```

Fftr.

src/layer/__tests__/outputResponses.ts
```diff
@@ -0,0 +1,73 @@
...
+  // arrayBuffer needs to be used, and removing this will cause getMap to fetch a blob, as window.Blob was created with jsdom
+  window.Blob = undefined;
+  const { layer, getMapParamsDefaultResponseId, mockedResponse } = constructFixtureGetMapOutputResponses();
+  setAuthToken(EXAMPLE_TOKEN);
+  mockNetwork.reset();
+  mockNetwork.onPost().replyOnce(200, mockedResponse);
+
+  await layer.getMap(getMapParamsDefaultResponseId, ApiType.PROCESSING);
+
+  expect(mockNetwork.history.post.length).toBe(1);
```
sinergise-anze, line 59: This is only checking that the request was made - a much better test would be if you could check that the parameter was set correctly, because it would guard against regressions.

Something like that:
```javascript
    const request = mockNetwork.history.post[0];
    const requestData = JSON.parse(request.data);
    expect(requestData.output.responses).toEqual({ identifier: "default", format: "image/jpeg" });
```
(I didn't test it)

I think this should be fixed (and below), but no need to assign back when it's done, just resolve and merge.

src/layer/__tests__/fixtures.outputResponses.ts
```diff
@@ -0,0 +1,48 @@
+import { S2L2ALayer, BBox, CRS_EPSG4326, MimeTypes } from '../../index';
```
sinergise-anze, line 1: IIUC, we don't actually need this fixture at all, because in the tests we don't care about the responses (we don't actually test the result), we just check that the correct request was made.

I would suggest removing it for clarity (the next person reading this will need to understand the same thing), but it's not a big deal. Fftr.

sinergise-anze (approved):

sinergise-anze (approved):

src/layer/processing.ts
zcernigoj: Thanks, kind of forgot to move the comment to a better place - moved it now to line 35 in `src/layer/const.ts` - much better location.
I also removed the check as suggested. Setting the empty string will hopefully never happen, but if it does, we can explain / change the behaviour then.

src/layer/processing.ts
zcernigoj: Thanks, totally overlooked that solution. Looks much nicer.

src/layer/__tests__/outputResponses.ts
zcernigoj: Thanks, I added tests like this. Will resolve and merge

src/layer/__tests__/fixtures.outputResponses.ts
zcernigoj: Thanks, yes, we don't care about responses (well, not yet, maybe we should add such test in the future). 
Removed this file and moved the needed things (`layer` and `getMapParams`) to `outputResponses.ts`.
Will resolve

src/layer/processing.ts
zcernigoj: Resolving

src/layer/processing.ts
zcernigoj: Resolving

PR #163, opened 2021-03-03
Image fetching optimization: automatically decide between JPEG and PNG

sinergise-anze:
As an optimization, we can decide for some layers (which don't output transparent values where the data is available) to use JPEG instead of PNG depending on availability of data. This PR adds `MimeTypes.JPEG_OR_PNG` "format", which causes this library to search for tiles and change format to JPEG (if the whole area is covered) or PNG (if not).

Further optimization is possible: we could search for the data availability for the whole screen to avoid excessive number of requests; however, this is something that can be improved in the next PR.

### DIFF
--- a/src/layer/AbstractDEMLayer.ts
+++ b/src/layer/AbstractDEMLayer.ts
@@ -71,13 +71,13 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
   }
 
   public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
-    const getMapValue = await ensureTimeout(async innerReqConfig => {
+    params = await this.decideJpegOrPng(params);
+    return await ensureTimeout(async innerReqConfig => {
       if (api === ApiType.PROCESSING) {
         await this.updateLayerFromServiceIfNeeded(innerReqConfig);
       }
       return await super.getMap(params, api, innerReqConfig);
     }, reqConfig);
-    return getMapValue;
   }
 
   protected async updateProcessingGetMapPayload(payload: ProcessingPayload): Promise<ProcessingPayload> {
--- a/src/layer/AbstractDEMLayer.ts
+++ b/src/layer/AbstractDEMLayer.ts
@@ -71,8 +71,8 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
   }
 
   public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
-    params = await this.decideJpegOrPng(params);
     return await ensureTimeout(async innerReqConfig => {
+      params = await this.decideJpegOrPng(params, innerReqConfig);
       if (api === ApiType.PROCESSING) {
         await this.updateLayerFromServiceIfNeeded(innerReqConfig);
       }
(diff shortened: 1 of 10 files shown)


src/layer/AbstractDEMLayer.ts
```diff
@@ -71,13 +71,13 @@ export class AbstractDEMLayer extends AbstractSentinelHubV3Layer {
   }
 
   public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
-    const getMapValue = await ensureTimeout(async innerReqConfig => {
+    params = await this.decideJpegOrPng(params);
```
Ardweaden, line 74: Shouldn't this also be inside the timeout?

src/layer/AbstractLayer.ts
```diff
@@ -201,19 +243,29 @@ export class AbstractLayer {
     bbox: BBox,
     fromTime: Date,
     toTime: Date,
-    maxFindTilesRequests: number = 50,
-    tilesPerRequest: number = 50,
+    maxFindTilesRequests: number | null = 50,
```
Ardweaden, line 246: Why is this needed? Isn't it already assigned a value if not passed?

src/layer/AbstractLayer.ts
```diff
@@ -201,19 +243,29 @@ export class AbstractLayer {
...
+    tilesPerRequest: number | null = 50,
     reqConfig?: RequestConfiguration,
+    overrideOrbitTimeMinutes: number | null = null,
```
Ardweaden, line 249: Why is this needed? (and set to infinity)

src/layer/AbstractLayer.ts
Ardweaden: I assume because you wanted to pass `overrideOrbitTimeMinutes` but didn't want to use `undefined`.

src/layer/AbstractDEMLayer.ts
sinergise-anze: Good point - yes, it should be, and it should also pass `innerReqConfig` along, to make sure timeout is observed. Thank you for pointing this out! Fixed.

src/layer/AbstractLayer.ts
sinergise-anze: Yes - in other places this library uses `null` as marker that some parameter was not specified and should be set to the default value. I am not sure if this was the best possible decision, but I used the same convention here for consistency.

Does that sound like a good idea or do you think a different approach would be better?

Ardweaden: I think this is an example why it's better to have a params object than separate params.

Ardweaden (approved):

src/layer/AbstractLayer.ts
sinergise-anze: Yes, but then you _must_ name all the parameters, every time. Unfortunately JS doesn't have a real "named params" mechanism like Python does.

sinergise-anze: When calling, you need to name the parameters:
```
  myFunction({ parameterA: 15, parameterB: 30 });
```

You can't just say:
```
  myFunction({15, 50});  // invalid syntax
```

Anyway... this is probably a discussion for the chat. :)

PR #164, opened 2021-03-10
Retry failed requests if rate limited

sinergise-anze:
This MR makes sh-js treat `429` responses the same as `5xx`. The reasoning behind it is that user doesn't really care about the reason for failure - if repeating the request might make it succeed, then it should be repeated.

Rate limiting responses can of course be treated a bit more subtly:
- failed requests (on the same page) could coordinate to avoid hitting the endpoint at the same time
- one 429 response could signal all other requests (to the same service) to delay a bit, thus avoiding multiple 429 responses

But this is left for the future MRs.

### DIFF
--- a/src/utils/axiosInterceptors.ts
+++ b/src/utils/axiosInterceptors.ts
@@ -129,5 +129,5 @@ const shouldRetry = (error: AxiosError): boolean => {
   if (!error.response || !error.response.status) {
     return false;
   }
-  return error.response.status >= 500 && error.response.status <= 599;
+  return error.response.status == 429 || (error.response.status >= 500 && error.response.status <= 599);
 };


zcernigoj (approved):
> ... user doesn't really care about the reason for failure ...

... if the reason is _not_ a wrongly formatted request. (maybe this would make the description better; or is it meant the user of the app that uses this library?)

Good to merge :+1:

sinergise-anze:
> > ... user doesn't really care about the reason for failure ...
> ... if the reason is _not_ a wrongly formatted request. (maybe this would make the description better; or is it meant the user of the app that uses this library?)

Sorry if it wasn't clear enough, what I meant is this:

> if repeating the request might make it succeed, then it should be repeated

With 400 repeating the request will not make it succeed, but with 429 this might happen (as it is a temporary failure).

Anyway, thanks for checking this, will merge!

PR #171, opened 2021-05-14
Fix identical request with a different authToken not being made after a failed one

Ardweaden:
After a request failed and finally rejected the promise, `cacheKey` wasn't removed from `cacheableRequestsInProgress`.

So making a request with identical parameters, but a different `authToken` (which isn't a part of the hash), didn't work.

### DIFF
--- a/src/layer/__tests__/auth.ts
+++ b/src/layer/__tests__/auth.ts
@@ -75,3 +75,19 @@ test('requestAuthToken correctly encodes URI parameters', async () => {
   const req = mockNetwork.history.post[0];
   expect(req.data).toBe('grant_type=client_credentials&client_id=asd%2C321&client_secret=.%2F*%26');
 });
+
+test('getMap with different authToken following an identical failed getMap makes a request', async () => {
+  const { layer, getMapParams } = constructFixtureGetMap();
+
+  mockNetwork.reset();
+  mockNetwork.onPost().replyOnce(429, '');
+  mockNetwork.onPost().replyOnce(200, '');
+
+  const reqConfig = { authToken: EXAMPLE_TOKEN1 };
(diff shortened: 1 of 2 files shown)


src/layer/__tests__/auth.ts
```diff
@@ -75,3 +75,19 @@ test('requestAuthToken correctly encodes URI parameters', async () => {
...
+  mockNetwork.onPost().replyOnce(429, '');
+  mockNetwork.onPost().replyOnce(200, '');
+
+  const reqConfig = { authToken: EXAMPLE_TOKEN1 };
+  await layer.getMap(getMapParams, ApiType.PROCESSING, reqConfig);
+
+  reqConfig.authToken = EXAMPLE_TOKEN2;
+  await layer.getMap(getMapParams, ApiType.PROCESSING, reqConfig);
+
+  expect(mockNetwork.history.post.length).toBe(2);
```
dgostencnik, line 92: Not sure what exactly is this test supposed to check. Should second request be made with `EXAMPLE_TOKEN2`?  (it is not). Perhaps it would be better to set appropriate caching and retrying settings to make things more clear.

Ardweaden: Yes, sorry, the test was incorrect. It should be more clear now.

dgostencnik: It is. I'd add test if last request was made with correct token. fftr
```
  expect(mockNetwork.history.post[3].headers.Authorization).toBe(`Bearer ${EXAMPLE_TOKEN2}`);
```

dgostencnik (approved):

src/layer/__tests__/auth.ts
Ardweaden: That's probably good to have. Added, together with a line which checks if one request earlier used the first token.

PR #172, opened 2021-05-20
Feature/add support for statistical api

dgostencnik:
Only most basic support for statisticalApi was added.  
- visibility of a few layer's properties and members had to be changed from protected to public
- FIS related code was extracted from layers and refactored to remove duplicated code 
- atm equal frequency histograms used by FIS are not supported by statisticalApi, similar result is achieved by using percentiles (more details in  1665 Statistical API: Equal frequency histogram)


src/statistics/StatisticalApiPayload.ts
```diff
@@ -0,0 +1,153 @@
...
+  aggregation: StatisticalApiAggregationPayload;
+  calculations: StatisticalApiCalculationsPayload;
+};
+
+export function createCalculationsPayload(
+  layer: AbstractSentinelHubV3Layer,
+  params: any,
+  output?: string,
+): StatisticalApiCalculationsPayload {
+  const outputId = output ? output : 'default';
```
Ardweaden, line 139: I guess this is good for the case where user would specify `histogram: equal_frequency` (so a legacy param), but we should provide a way for the user to pass custom `calculations`.

src/statistics/StatisticalApi.ts
```diff
@@ -0,0 +1,126 @@
...
+      `${layer.getShServiceHostname()}api/v1/statistics`,
+      {
+        input: input,
+        aggregation: aggregation,
+        calculations: calculations,
+      },
+      requestConfig,
+    );
+
+    return this.convertToFISResponse(response.data, STATS_DEFAULT_OUTPUT);
```
Ardweaden, line 124: I am not convinced this is the best idea. IMO, this should be a utility function, and Statistical API should return the results in the new format.

That is, if someone wanted to switch to Stat API from FIS, they should change the statistics provider, and then use this utility function on the output, if they wanted to use the same format.

Ardweaden (changes requested):
I am not sure we should be too concerned with compatibility with soon-to-be-deprecated FIS. That is, I don't think switching between the two statistics providers should be supported in both ways. It's great that switching from FIS to stat API was made relatively seamless, but I don't think the interface should be identical. IMO we should support stat API as stat API, including all the extra parameters it needs and having a different output.

Because long-term (when FIS gets deprecated, although I know you are quite sceptical how quickly that will actually happen) it would probably bring nothing but trouble to have some quirky artefacts which would stem from the way FIS worked...

src/statistics/StatisticalApiPayload.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  if (!params.toTime) {
+    throw new Error('toTime must be defined');
+  }
+
+  if (!params.aggregationInterval) {
+    throw new Error('aggregationInterval must be defined');
+  }
+  //TODO: properly handle resolution
+  const resX = params.resolution;
```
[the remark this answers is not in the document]
Ardweaden: I guess we don't need to handle this now.

src/statistics/StatisticalApi.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!authToken) {
+      throw new Error('Must be authenticated to use Statistical API');
+    }
+
+    await layer.updateLayerFromServiceIfNeeded(reqConfig);
+
+    const input = await StatisticsUtils.createInputPayload(layer, params, reqConfig);
+    const aggregation = StatisticsUtils.createAggregationPayload(layer, {
+      ...params,
+      aggregationInterval: 'P1D',
```
Ardweaden, line 31: If I understand this correctly, the `aggregationInterval` is hardcoded to one day? 

I guess it's indeed best  if we just implement what we need for EOB and then we can deal with full support of all Statistical API parameters later (if required).

Ardweaden (approved):

PR #175, opened 2021-06-04
Add CRS:84 as a separate CRS from EPSG:4326

Ardweaden:
`CRS:84` is not identical to `EPSG:4326`, they have switched coordinates (in some cases).

So when requesting FIS with  `EPSG:4326`, but using GeoJSON (which has `CRS:84`), the service switched the coordinates around.

### DIFF
--- a/src/crs.ts
+++ b/src/crs.ts
@@ -35,7 +35,13 @@ export const CRS_EPSG4326: CRS = {
   opengisUrl: 'http://www.opengis.net/def/crs/EPSG/0/4326',
 };
 
-export const CRS_WGS84: CRS = CRS_EPSG4326;
+export const CRS_WGS84: CRS = {
+  authId: 'CRS:84',
+  auth: 'CRS',
+  srid: 84,
+  urn: 'urn:ogc:def:crs:OGC:1.3:CRS84',
+  opengisUrl: 'http://www.opengis.net/def/crs/OGC/1.3/CRS84',
+};
 
 export const SUPPORTED_CRS_OBJ = {
   [CRS_EPSG3857.authId]: CRS_EPSG3857,
--- a/src/crs.ts
+++ b/src/crs.ts
@@ -35,6 +35,9 @@ export const CRS_EPSG4326: CRS = {
   opengisUrl: 'http://www.opengis.net/def/crs/EPSG/0/4326',
 };
 
+/**
+ * Same as EPSG:4326, but with a switched coordinate order.
+ */
 export const CRS_WGS84: CRS = {
   authId: 'CRS:84',
   auth: 'CRS',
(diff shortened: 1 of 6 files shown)


src/crs.ts
```diff
@@ -35,7 +35,13 @@ export const CRS_EPSG4326: CRS = {
   opengisUrl: 'http://www.opengis.net/def/crs/EPSG/0/4326',
 };
 
-export const CRS_WGS84: CRS = CRS_EPSG4326;
+export const CRS_WGS84: CRS = {
```
zcernigoj, line 38: Should we add a comment that this is the same as WGS84 / EPSG:4326, only with flipped coordinates?

zcernigoj:
The change in `src/layer/AbstractSentinelHubV3Layer.ts` should probably be done in `src/layer/AbstractSentinelHubV1OrV2Layer.ts` too.

zcernigoj:
I think there should be a test for the `getStats()` function which would capture the request that the function creates and check the order of the coordinates for different CRSs.

But maybe this can be added later.

Ardweaden:
> The change in `src/layer/AbstractSentinelHubV3Layer.ts` should probably be done in `src/layer/AbstractSentinelHubV1OrV2Layer.ts` too.

@zcernigoj I don't seem to be able to reply directly?

Anyway, thanks, added.

Ardweaden:
> I think there should be a test for the `getStats()` function which would capture the request that the function creates and check the order of the coordinates of the geometry for different CRSs.
> 
> But maybe this can be added later.

@zcernigoj I don't think there's any interesting tests that could be added, except for checking if the specified CRS is in the payload. We don't change the order of coordinates in the FIS request.

Therefore, I instead added 2 stories, one for V3 and one for V12.

src/crs.ts
Ardweaden: Sure.

zcernigoj: Thanks. Resolving

zcernigoj:
> > The change in `src/layer/AbstractSentinelHubV3Layer.ts` should probably be done in `src/layer/AbstractSentinelHubV1OrV2Layer.ts` too.
> 
> @zcernigoj I don't seem to be able to reply directly?
> 
> Anyway, thanks, added.

I think Github doesn't that functionality for the comments that are not in the code. 

> > I think there should be a test for the `getStats()` function which would capture the request that the function creates and check the order of the coordinates of the geometry for different CRSs.
> > But maybe this can be added later.
> 
> @zcernigoj I don't think there's any interesting tests that could be added, except for checking if the specified CRS is in the payload. We don't change the order of coordinates in the FIS request.
> 
> Therefore, I instead added 2 stories, one for V3 and one for V12.

:+1:

zcernigoj (approved):
Can be merged :+1:

Ardweaden:
> I think Github doesn't that functionality for the comments that are not in the code.

@zcernigoj worse UI than EOB search time range.

PR #176, opened 2021-06-07
Add Landsat preview image links

dthiex:
For Landsat we didn't jet add preview images. This PR adds preview image links
- for L4-5 L1
- L4-5 L2 (still needs to be done)
- L8 L2 (still needs to be done)


dthiex:
Closing because as described in https://github.com/sentinel-hub/sentinelhub-js/issues/207 this is to complicated for what it brings.

Issue #174, opened 2021-06-02
Cancelled requests don't get removed from `cacheableRequestsInProgress`


Ardweaden:
When a request gets cancelled, the `Cancel` error doesn't contain `config` with would have the `cacheKey`, so the request is never removed from `cacheableRequestsInProgress`.

dgostencnik:
should be fixed with PR #178

PR #178, opened 2021-06-18
remove cancelled requests from list of requests in progress

dgostencnik:
As described in #174, after canceling request was not removed from the list of requestsInProgress which result in infinite wait for cache lock when the same request was made next time.

What was done
- cacheKey was added to our CancelToken and it is set by axios interceptor when trying to get cache lock (fetchCachedResponse)
- cackeKey is removed from cacheableRequestsInProgress when cancel is invoked

The problem can be observed in Sentinel 2 L1C˙ stories Cancel Requests example` where repeating request after cancellation resulted in the broken image and missing requests.

Previous behavior can be replicated by commenting out code code added to the interceptor or cancel() method and running `cancelToken` tests (2 should fail) or story mentioned above.

### DIFF
--- a/src/utils/cacheHandlers.ts
+++ b/src/utils/cacheHandlers.ts
@@ -15,7 +15,7 @@ export type CacheConfig = {
 // still all be executed - because by the time first response is saved in cache, the other 9
 // requests are already made too. To combat this, we save cacheKeys of ongoing requests and
 // simply delay new requests with the same cacheKey.
-const cacheableRequestsInProgress = new Set();
+export const cacheableRequestsInProgress = new Set();
 
 export const removeCacheableRequestsInProgress = (cacheKey: string): void => {
   cacheableRequestsInProgress.delete(cacheKey);
@@ -30,7 +30,6 @@ export const fetchCachedResponse = async (request: any): Promise<any> => {
   if (cacheKey === null) {
     return request;
   }
-
   if (request.cancelToken && request.setCancelTokenCacheKey) {
     request.setCancelTokenCacheKey(cacheKey);
   }
--- a/src/utils/cacheHandlers.ts
+++ b/src/utils/cacheHandlers.ts
@@ -30,6 +30,9 @@ export const fetchCachedResponse = async (request: any): Promise<any> => {
   if (cacheKey === null) {
     return request;
   }
+
+  // When request is cancelled, it must also be removed from list of cacheableRequestsInProgress.
+  // In order to remove requests from cacheableRequestsInProgress, cancelToken must be aware of requests(cacheKeys) it is responsible for.
   if (request.cancelToken && request.setCancelTokenCacheKey) {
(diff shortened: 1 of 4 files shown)


src/utils/cancelRequests.ts
```diff
@@ -14,13 +14,22 @@ export type RequestConfiguration = {
...
   protected source: CancelTokenSource | null = null;
+  protected cacheKey: string | null = null;
 
   public constructor() {
     this.source = axios.CancelToken.source();
     this.token = this.source.token;
   }
 
+  public setCancelTokenCacheKey(cacheKey: string | null): void {
+    this.cacheKey = cacheKey;
```
Ardweaden, line 25: A single `CancelToken` can be used for multiple requests. So I think this should be an array with all the `cacheKeys` inside, and then all would be removed on cancelling.

I believe this fails:

```
it('handles multiple requests with the same cancel token', async () => {
    let requestError = null;
    const { requestPromise, thenFn, catchFn, cancelToken, mockedResponse } = createRequestPromise(
      true,
      err => (requestError = err),
    );

    const {layer, getMapParams, mockedResponse: mockedResponse2} = constructFixtureGetMap()

    mockNetwork.onPost().replyOnce(200, mockedResponse);
    mockNetwork.onPost().replyOnce(200, mockedResponse2);

    setAuthToken("EXAMPLE_TOKEN");

    await Promise.all([requestPromise, layer.getMap(getMapParams, ApiType.PROCESSING, {cancelToken: cancelToken, cache: {
      expiresIn: 60,
      targets: [CacheTarget.MEMORY],
    }}).catch((err:any) => console.log(err)), setTimeout(() => cancelToken.cancel(), 10)]);

    expect(cacheableRequestsInProgress.size).toBe(0);
  });
  ```

File: src/utils/cacheHandlers.ts [hunk 2]
```diff
@@ -30,6 +30,9 @@ export const fetchCachedResponse = async (request: any): Promise<any> => {
   if (cacheKey === null) {
     return request;
   }
+  if (request.cancelToken && request.setCancelTokenCacheKey) {
```
Ardweaden, line 33: I think we will have to add a comment here, as it might be pretty confusing why this is here if the dev isn't aware of the bug.

Ardweaden (changes requested):
I have to leave a comment, as I didn't create the other comments as a part of a review.

src/utils/cancelRequests.ts
dgostencnik: Thanks for spotting this, I didn't think about that. I added test and changed to support multiple cacheKeys.

src/utils/cacheHandlers.ts
dgostencnik: Some comment was added, but whole thing is still confusing

Ardweaden (approved):

PR #185, opened 2021-07-23
Set legend in updateLayerFromServiceIfNeeded in BYOCLayers.

Ardweaden:
Setting legend was missing.

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -81,7 +81,9 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       if (this.collectionId === null || this.evalscript === null || this.subType === null) {
         const layerParams = await this.fetchLayerParamsFromSHServiceV3(innerReqConfig);
         this.collectionId = layerParams['collectionId'];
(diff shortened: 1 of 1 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -81,6 +81,7 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       if (this.collectionId === null || this.evalscript === null || this.subType === null) {
         const layerParams = await this.fetchLayerParamsFromSHServiceV3(innerReqConfig);
         this.collectionId = layerParams['collectionId'];
+        this.legend = layerParams['legend'] ? layerParams['legend'] : null;
```
dgostencnik, line 84: I would follow the pattern and check if value for `legend` has already been set on layer.  Not sure if really needed, fftr 
```javascript
if (!this.legend){
}
```

dgostencnik (approved):

src/layer/BYOCLayer.ts
Ardweaden: Done that.

PR #186, opened 2021-08-10
Feature/new landsat datasets

dgostencnik:
3 new Landsat layers were added 
- Landsat15AWSLMSSL1Layer : Landsat 1-5 MSS L1 from 1972 to 1993
- Landsat7AWSLETML1Layer and Landsat7AWSLETML2Layer: Landsat 7 ETM+ L1 and L2 (from 1999 to May 2003, horizontal lines afterwards)


Ardweaden (commented):
All three have a filtering option `tiers` (https://docs.sentinel-hub.com/api/latest/data/landsat-mss/#tiers).

I think this should be added, except if there is some reason for omitting this.

dgostencnik:
Actually `tiers` filtering option has been added for all Landsat products recently. I created an issue for this and it will be added as separate PR, because I'm 100% sure I'll mess something up in `updateFromServiceIfNeeded`

Ardweaden (approved):

PR #188, opened 2021-08-17
Add speckle filter for Sentinel 1

mitjak:


src/layer/S1GRDAWSEULayer.ts
```diff
@@ -122,6 +137,7 @@ export class S1GRDAWSEULayer extends AbstractSentinelHubV3Layer {
       this.orthorectify = layerParams['orthorectify'];
       this.demInstanceType = layerParams['demInstance'];
       this.orbitDirection = layerParams['orbitDirection'] ? layerParams['orbitDirection'] : null;
+      this.speckleFilter = layerParams['speckleFilter'];
```
Ardweaden, line 140: This - and it should be fixed for other params as well - shouldn't override the existing values, if they are set.

This is most likely the reason why it currently doesn't work in EOB.

src/layer/S1GRDAWSEULayer.ts
```diff
@@ -38,6 +38,18 @@ export enum Resolution {
   MEDIUM = 'MEDIUM',
 }
 
+export enum SpeckleFilterType {
+  DEFAULT = 'DEFAULT',
```
Ardweaden, line 42: According to the documentation, `DEFAULT` is not one of the possible values of speckle filter type.

Reminder, this library is intended to provide a JS interface for using our APIs, similar to `sentinelhub-py` for Python (https://docs.sentinel-hub.com/api/latest/reference/#operation/process). 

EOB and most other apps of ours happen to use it, but it's made so that it can be used by anyone for any purpose, so it shouldn't have any knowledge of those apps.

Ardweaden (approved):

PR #156, opened 2021-02-03
Allow users to dynamically rewrite any request URL with callback function

sinergise-anze:
This adds `rewriteUrlFunc` to `requestsConfig`.

I didn't add documentation for it as I'm not 100% sure that this should be public. :)

The storybook for GIBS uses this functionality to redirect `GetCapabilities` requests to another URL.

PR #189, opened 2021-08-27
Support zarr and staging services.

Ardweaden:
Not sure if we should list the staging services among the locations. Perhaps a better idea would be to pass an extra param `baseUrl` or `locationUrl` which overrides anything if present.


zcernigoj:
Would what is done in https://github.com/sentinel-hub/sentinelhub-js/pull/156 be of any help in this situation?

zcernigoj (approved):
good to go :+1:

PR #192, opened 2021-09-27
Fix/aux request cache

### LINKED ISSUE #180: Caching of auxiliary requests
https://git.sinergise.com/team-6/eobrowser3/-/merge_requests/598#note_216491

When creating a layer, the layer usually doesn't have mosaickingOrder, upsampling and downsampling set. We do `updateLayerFromServiceIfNeeded`, but it doesn't set either. The `/layers` and getCapabilities responses are saved in the memory. 

However, when we make a getMap request, the method _again_ calls `/layers`. Because of this line: https://github.com/sentinel-hub/sentinelhub-js/blob/9cd7853478f79cf444786d6f2fcf5c5f525d380d/src/layer/AbstractSentinelHubV3Layer.ts#L197

However, because we use storage cache for `getMap`, it reads the `/layers` response from *storage*. However, the one saved there is from when we initially loaded the instance, without the new layer.

---------

I think the latest assumption is that we should only cache `getCapabilties/layer` to memory, but as you found out `fetchLayerParamsFromSHServiceV3` is cached to `cacheApi` because of the `reqConfig` passed from `getMap`, so I think we should just hardcode caching to memory for `fetchLayerParamsFromSHServiceV3`, this should be ok as it is a protected method.

The following changes should work.

```
 protected async fetchLayerParamsFromSHServiceV3(reqConfig: RequestConfiguration): Promise<any> {
...
    const requestConfig: AxiosRequestConfig = {
      responseType: 'json',
      headers: headers,
      ...getAxiosReqParams(reqConfig, { ...CACHE_CONFIG_30MIN, targets: [CacheTarget.MEMORY] }),
    };<
```

```
export const getAxiosReqParams = (
  reqConfig: RequestConfiguration,
  defaultCache: CacheConfig,
): AxiosRequestConfig => {
...

  if (reqConfigWithDefault.cache && reqConfigWithDefault.cache.expiresIn) {
    axiosReqConfig.cache.expiresIn = reqConfigWithDefault.cache.expiresIn;
  }
  if (reqConfigWithDefault.cache && reqConfigWithDefault.cache.targets) {
    axiosReqConfig.cache.targets = reqConfigWithDefault.cache.targets;
  }
```

------

Is there really a point in passing the `getMap`s `reqConfig` to `fetchLayerParamsFromSHServiceV3` at all, though? If we are deciding to cache by default for the user, we could literally just have it hardcoded to use memory regardless.

guss84:
resolve #180 

- Introduces `CACHE_CONFIG_30MIN_MEMORY`
- Removes cache config received from `getMap` in `fetchLayerParamsFromSHServiceV3` and assign  `CACHE_CONFIG_30MIN_MEMORY` to the cache property of `reqConfig`

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -114,7 +114,8 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     // we do not want this as layer updates will not invalidate this cache, so we rathger cache to memory
     const reqConfigWithMemoryCache = {
       ...reqConfig,
-      cache: CACHE_CONFIG_30MIN_MEMORY,
+      // Do not override cache if cache is disabled with `expiresIn: 0`
+      cache: reqConfig.cache.expiresIn === 0 ? reqConfig.cache : CACHE_CONFIG_30MIN_MEMORY,
     };
     const requestConfig: AxiosRequestConfig = {
       responseType: 'json',
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -31,8 +31,7 @@ import { ensureTimeout } from '../utils/ensureTimeout';
 
 import { Effects } from '../mapDataManipulation/const';
 import { runEffectFunctions } from '../mapDataManipulation/runEffectFunctions';
-import { CACHE_CONFIG_30MIN, CACHE_CONFIG_NOCACHE } from '../utils/cacheHandlers';
-
+import { CACHE_CONFIG_30MIN, CACHE_CONFIG_30MIN_MEMORY, CACHE_CONFIG_NOCACHE } from '../utils/cacheHandlers';
 interface ConstructorParameters {
   instanceId?: string | null;
   layerId?: string | null;
@@ -110,10 +109,17 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const headers = {
       Authorization: `Bearer ${authToken}`,
     };
+
+    // reqConfig might include the cache config from getMap, which could cache instances/${this.instanceId}/layers
+    // we do not want this as layer updates will not invalidate this cache, so we rathger cache to memory
+    const reqConfigWithMemoryCache = {
+      ...reqConfig,
+      cache: CACHE_CONFIG_30MIN_MEMORY,
+    };
     const requestConfig: AxiosRequestConfig = {
       responseType: 'json',
       headers: headers,
-      ...getAxiosReqParams(reqConfig, CACHE_CONFIG_30MIN),
+      ...getAxiosReqParams(reqConfigWithMemoryCache, null),
     };
     const res = await axios.get(url, requestConfig);
     const layersParams = res.data.map((l: any) => ({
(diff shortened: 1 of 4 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -110,10 +109,17 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     const headers = {
       Authorization: `Bearer ${authToken}`,
     };
+
+    // reqConfig might include the cache config from getMap, which could cache instances/${this.instanceId}/layers
+    // we do not want this as layer updates will not invalidate this cache, so we rathger cache to memory
+    const reqConfigWithMemoryCache = {
+      ...reqConfig,
+      cache: CACHE_CONFIG_30MIN_MEMORY,
```
dgostencnik, line 117: What if caching is disabled in `reqConfig` by setting `expiresIn` to 0 ? This will now override that setting and cache it in memory for 30mins. Not sure if this is expected behaviour.

dgostencnik (commented):
It would probably make sense to write tests for this, but it might be quite a challenging task. I'm not sure if we can tell easily from where the cache is being served.

src/layer/AbstractSentinelHubV3Layer.ts
guss84: Good point, and I am not sure what the best approach here would be. The information needed from `fetchLayerParamsFromSHServiceV3` is stored in the layer's instance, so `fetchLayerParamsFromSHServiceV3` could only called if a new instance is created.

Do you think removing cache here is a viable option, or should we only override cache if cache was set in `reqConfig` unless it is `expiresIn: 0`?

dgostencnik: Yes, I think if cache is disabled for request, I wouldn't override it in `fetchLayerParamsFromSHServiceV3`.

guss84:
> It would probably make sense to write tests for this, but it might be quite a challenging task. I'm not sure if we can tell easily from where the cache is being served.

I have added 2 test cases, one for checking if it is cached to memory by default and the other to test that caching is disabled.

src/layer/AbstractSentinelHubV3Layer.ts
guss84: Made the changes so that `fetchLayerParamsFromSHServiceV3` will not be cached if caching is disabled by setting `expiresIn: 0`

src/layer/__tests__/cache.ts
```diff
@@ -496,3 +497,56 @@ describe('Unit test for cacheStillValid', () => {
...
+describe('Unit test for aux request caching', () => {
+  beforeEach(async () => {
+    Object.assign(global, makeServiceWorkerEnv(), fetch); // adds these functions to the global object
+    await invalidateCaches();
+    setAuthToken(undefined);
+  });
+  it('It should be cache aux request to memory by default', async () => {
+    setAuthToken(EXAMPLE_TOKEN);
+    const { layer, mockedResponse, expectedLayerParams } = constructFixtureUpdateLayerFromServiceIfNeeded({});
+    const requestsConfig = {
```
dgostencnik, line 510: You could expand this test a bit by using  it.each and use different requestsConfigs as an input
listOfRequstConfigs=[
{} , 
null, 
undefined, 
{ retries: 1 },
{
      cache: {
        expiresIn: 60,
      },
   }
... all other combination of expiresIn and target params,
]
  
 it.each([....listOfRequstConfigs])('It should be cache aux request to memory by default', async requestsConfig => {....})

fftr

dgostencnik (approved):

src/layer/__tests__/cache.ts
guss84: I expanded the test as suggested and added other `requestsConfigs`

PR #194, opened 2021-12-08
Feature/wmts

guss84:
- Support for WMTS and Planet NICFI


src/layer/LayersFactory.ts
```diff
@@ -293,4 +314,73 @@ export class LayersFactory {
...
+    reqConfig: RequestConfiguration,
+  ): Promise<AbstractLayer[]> {
+    let newLayers = [];
+    const parsedLayers = await fetchLayersFromWmtsGetCapabilitiesXml(baseUrl, reqConfig);
+    for (let layerInfo of parsedLayers) {
+      const layer = {
+        layerId: layerInfo.Name[0],
+        title: layerInfo.Title[0],
+        description: layerInfo.Abstract ? layerInfo.Abstract[0] : null,
+        dataset: null as string | null,
```
Ardweaden, line 361: I don't quite understand what this is supposed to do tbh? 

Any way, we usually make a dataset object for collection-specific layers. I think it wouldn't hurt making it here, although admittedly there probably isn't much benefit to it. Just to be consistent.

src/layer/WmtsLayer.ts
```diff
@@ -0,0 +1,104 @@
...
+        const layer = parsedLayers.find(layerInfo => this.layerId === layerInfo.Name[0]);
+        this.resourceUrl = layer.ResourceUrl;
+      }
+    }, reqConfig);
+  }
+
+  public async getMap(params: GetMapParams, api: ApiType, reqConfig?: RequestConfiguration): Promise<Blob> {
+    return await ensureTimeout(async innerReqConfig => {
+      await this.updateLayerFromServiceIfNeeded(reqConfig);
+      const paramsWithoutEffects = { ...params };
```
Ardweaden, line 52: Why are we removing effects and never applying them?

src/layer/WmtsLayer.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+  }
+
+  public getMapUrl(params: GetMapParams, api: ApiType): string {
+    if (api !== ApiType.WMTS) {
+      throw new Error('Only WMTS is supported on this layer');
+    }
+
+    if (!params.bbox && !params.tileCoord) {
+      throw new Error('No bbox or x,y coordinates provided');
```
Ardweaden, line 74: Minor: I assume `z` is missing.

src/layer/WmtsLayer.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+    if (!this.resourceUrl) {
```
Ardweaden, line 76: If this is not available, we could make the `updateLayerFromServiceIfNeeded` request first, like it's done elsewhere. I'm not a big fan of that because it only leads to bugs imho. Just bringing it up to acknowledge it.

src/layer/wmts.utils.ts
```diff
@@ -0,0 +1,124 @@
...
+
+function getZoomFromBbox(bbox: BBox): number {
+  if (bbox.crs === CRS_EPSG4326) {
+    const topLeft = toMercator([bbox.minX, bbox.minY]);
+    const bottomRight = toMercator([bbox.maxX, bbox.maxY]);
+    bbox = new BBox(CRS_EPSG3857, topLeft[0], topLeft[1], bottomRight[0], bottomRight[1]);
+  }
+  return Math.max(
+    0,
+    Math.min(19, 1 + Math.floor(Math.log(EQUATOR_LEN / ((bbox.maxX - bbox.minX) * 1.001)) / Math.log(2))),
```
Ardweaden, line 62: I don't really understand this function fully. Does it assume the bbox exactly matches a tile?

src/layer/WmtsLayer.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.baseUrl = baseUrl;
+    this.layerId = layerId;
+    this.resourceUrl = resourceUrl;
+  }
+
+  public async updateLayerFromServiceIfNeeded(reqConfig?: RequestConfiguration): Promise<void> {
+    await ensureTimeout(async innerReqConfig => {
+      if (!this.resourceUrl) {
+        const parsedLayers = await fetchLayersFromWmtsGetCapabilitiesXml(this.baseUrl, innerReqConfig);
+        const layer = parsedLayers.find(layerInfo => this.layerId === layerInfo.Name[0]);
```
Ardweaden, line 43: Will this work for Planet layers with "fake" `layerId`s?

src/layer/WmtsLayer.ts
```diff
@@ -0,0 +1,104 @@
...
+interface ConstructorParameters {
+  baseUrl?: string;
+  layerId?: string;
+  title?: string | null;
+  description?: string | null;
+  legendUrl?: string | null;
+  resourceUrl?: string | null;
+}
+
+export class WmtsLayer extends AbstractLayer {
```
Ardweaden, line 20: WMTS layer should implement `supportsApiType`.

src/layer/WmtsLayer.ts
guss84: No it will not, but only `makePlanetBasemapLayers` creates theses "fake" layers, and sets `resourceUrl` when constructing the layer.

src/layer/WmtsLayer.ts
guss84: This would require us to change `AbstractLayer.getMapUrl` to an async function, which would be a braking change, and not really worth it in my eyes.

src/layer/LayersFactory.ts
```diff
@@ -293,4 +314,73 @@ export class LayersFactory {
...
+    const filteredLayersInfos =
+      filterLayers === null ? layersInfos : layersInfos.filter(l => filterLayers(l.layerId, l.dataset));
+
+    return filteredLayersInfos.map(
+      ({ layerId, title, description, legendUrl, resourceUrl }) =>
+        new WmtsLayer({ baseUrl, layerId, title, description, legendUrl, resourceUrl }),
+    );
+  }
+
+  // Analtyc layers accept a proc parameter to specify a dynamically-rendered false color visualization, for example NDVI.
```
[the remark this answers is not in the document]
guss84: fixed

src/layer/LayersFactory.ts
guss84: Added a dataset object

src/layer/WmtsLayer.ts
guss84: Thanks, didnt notice I wasnt adding the effects function

src/layer/WmtsLayer.ts
guss84: fixed

src/layer/wmts.utils.ts
guss84: Currently we can pass a the bbox of a tile, and get the correct X,Y,Z of that tile from a bbox, this will later be extended to support arbitrary bboxes

src/layer/LayersFactory.ts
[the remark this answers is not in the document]
Ardweaden: I didn't bother making issues for other typos haha

src/layer/WmtsLayer.ts
Ardweaden: Why not making an `updateLayerFromServiceIfNeeded` method on the Planet layer then?

guss84: Currently `updateLayerFromServiceIfNeeded` only sets `resourceUrl` , and it is a method of `WmtsLayer` because `resourceUrl`is a WMTS standard.

Ardweaden (approved):

PR #195, opened 2021-12-20
Fix/planet nicfi force rgb image

guss84:
planet_medres_normalized_analytic_2017-06_2017-11_mosaic should return an RGB image, but it returns a false colour color image.

This MR adds a `proc=rgb` to `planet_medres_normalized_analytic` layers in order to force an RGB image.


dgostencnik:
As far as I can tell, both `makePlanetBasemapLayers` and `findDatesUTC` should work correctly, but  I think both functions would benefit if a test or two was added (can be done on some other occasion).

PR #198, opened 2022-01-12
Fix/planet env storybook

### LINKED ISSUE #197: PLANET_API_KEY "placeholder" missing in .env.example
When running the storybook, the `PLANET_API_KEY` is mandatory, but the "placeholder" for it isn't present in the `.env.example` file, suggesting it's not mandatory.

guss84:
fixes #197 

Also noticed that one of the stories had a bug caused by an api change to the `tileCoord` object


stories/wmts.planet.stories.js
```diff
@@ -95,7 +94,6 @@ export const getMapWTmsLayersFactory = () => {
       width: 256,
```
zcernigoj, line 94: There's probably a typo in the name of this story - `getMapWTmsLayersFactory` -> should be `getMapWmtsLayersFactory`

stories/wmts.planet.stories.js
```diff
@@ -37,7 +37,6 @@ export const getMapBbox = () => {
...
       height: 256,
       format: MimeTypes.JPEG,
-      zoom: 14,
```
zcernigoj, line 40: If I understand this correctly, there are 2 ways of setting the "location" 
- bbox (and no zoom is required)
- tileCoord

The **bbox and no zoom** conclusion is based on no `zoom` being present in `src/layer/WmtsLayer.ts`.
If this conclusion is correct, should the `zoom` also be removed from:
- `GetMapParams` in `src/layer/const.ts`
- `WmtsLayer.getMap uses bbox` test in `src/layer/__tests__/WmtsLayer.ts`
-

stories/wmts.planet.stories.js
guss84: Fixed, thanks

stories/wmts.planet.stories.js
guss84: Fixed. Thanks for the thorough check.

stories/wmts.planet.stories.js
zcernigoj: :+1: Resolving

stories/wmts.planet.stories.js
zcernigoj: Looks good now. Resolving

zcernigoj (approved):

PR #202, opened 2022-01-31
Feature/stitch wmts images

guss84:
PR enables creating a single image from multiple WMTS tiles when the bbox from the `getMap` params require multiple tiles to be stitched . 

- Stitch images when bbox is not null, fetch single image when `bbox` is null and `tileCoords` are provided
- Introduce `matrixSet` to WMTS class, in order to get tile width, height,  zoom levels, and column and row count at each zoom

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -104,7 +104,7 @@ function parseUrl(
 }
 
 if (typeof window.URL.createObjectURL === 'undefined') {
-  window.URL.createObjectURL = (): string => 'kjkj';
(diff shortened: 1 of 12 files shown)


jest-setup.ts
```diff
@@ -102,4 +103,22 @@ function parseUrl(
   };
 }
 
+if (typeof window.URL.createObjectURL === 'undefined') {
+  window.URL.createObjectURL = (): string => 'kjkj';
```
mitjak, line 107: `kjkj`?

guss84: Was just a random string for this mocking. I have changed the string in hope that it clarifies it better.

PR #209, opened 2022-03-30
Use layers endpoint instead of getCapabilities in makeLayers

### LINKED ISSUE #208: Use `layers` endpoint instead of getCapabilities in makeLayers
Atm `makeLayers` uses getCapabilities to create layers. For `AbstractSentinelHubV3Layer` this could be replaced by using `layer` endpoint https://docs.sentinel-hub.com/api/latest/reference/#tag/configuration_layer

Main reason for implementing this is 
- getCapabilities doesn't provide all necessary layer data and we need to make requests to `layers` to get additional info
- we would like to disable ogc requests in playground for public instances  

It seems that adding parameter `preferGetCapabilities` which is true by default and use this parameter to switch between  `getCapabilities` and `layers` in `makeLayersSHv3` shouldn't introduce any breaking changes.

dgostencnik:
closes #208 

- additional parameter `preferGetCapabilities: boolean = true` was added to `LayersFactory.makeLayers`  and `legacyGetMapFromParams`
- code for fetching  layer parameters from configuration service was moved from `AbstractSentinelHubV3Layer` to utils 
- some tests for `makeLayers` were added


Ardweaden (approved):
I don't have anything to comment on, the changes seem logical and straightforward.

PR #210, opened 2022-03-31
fix types for `mimeFormat` parameter for `canvasToBlob()` function 

zcernigoj:
~~Export `MimeType` so it can be used in `canvasToBlob` which accepts `MimeType` not (already exported) `MimeTypes`~~
This isn't the best solution because it clashes with the JS built-in type MimeType if you don't add alias for it in the import statement.

Thus I decided to add the `FormatJpegOrPng` as a possible type for `mimeFormat` for `canvasToBlob()` function so predefined values of `MimeTypes` can be used.
If `MimeType.JPEG_OR_PNG` is passed to `canvasToBlob()` an error is thrown.

### DIFF
--- a/src/utils/canvas.ts
+++ b/src/utils/canvas.ts
@@ -1,4 +1,4 @@
-import { MimeType, ImageProperties } from '../layer/const';
+import { MimeType, FormatJpegOrPng, MimeTypes, ImageProperties } from '../layer/const';
 
 export async function drawBlobOnCanvas(
   ctx: CanvasRenderingContext2D,
@@ -21,7 +21,13 @@ export async function drawBlobOnCanvas(
   }
 }
 
-export async function canvasToBlob(canvas: HTMLCanvasElement, mimeFormat: MimeType): Promise<Blob> {
+export async function canvasToBlob(
+  canvas: HTMLCanvasElement,
+  mimeFormat: MimeType | FormatJpegOrPng,
+): Promise<Blob> {
+  if (mimeFormat === MimeTypes.JPEG_OR_PNG) {
+    return Promise.reject('This format is not supported');
+  }
   return await new Promise(resolve => canvas.toBlob(resolve, mimeFormat));
 }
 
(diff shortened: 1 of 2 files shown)


src/utils/canvas.ts
```diff
@@ -21,7 +21,13 @@ export async function drawBlobOnCanvas(
   }
 }
 
-export async function canvasToBlob(canvas: HTMLCanvasElement, mimeFormat: MimeType): Promise<Blob> {
+export async function canvasToBlob(
+  canvas: HTMLCanvasElement,
+  mimeFormat: MimeType | FormatJpegOrPng,
+): Promise<Blob> {
+  if (mimeFormat === MimeTypes.JPEG_OR_PNG) {
```
dgostencnik, line 28: I don't understand this. Why was it added to supported types if it is not supported? wouldn't it make more sense to 
- use correct MimeType instead of FormatJpegOrPng
- use mimeFormat: MimeType|string
- use canvas.toBlob instead of canvasToBlob wrapper

I can'

dgostencnik (commented):
This makes no sense to me, but I guess it can be merged if it was tested and/or  if it works or solves whatever it is supposed to fix.

src/utils/canvas.ts
zcernigoj: I agree it was a mess. I now went with the second option, _use mimeFormat: MimeType|string_ 

> use canvas.toBlob instead of canvasToBlob wrapper

can be used anyway already, but it's annoying because you have to pass the callback function to it to use the generated blob ([docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob)).
With `canvasToBlob()` you can simply do  `const blob = await canvasToBlob(...)`

And I guess we don't have to worry about the unsupported formats. From the docs:
> The default type is `image/png`; that type is also used if the given type isn't supported.

dgostencnik (approved):

PR #212, opened 2022-04-05
Temporarily partially handle ZARR collections

Ardweaden:
ZARR collections don't work atm because of https://git.sinergise.com/sentinel-core/java/-/issues/1985

This helps making it work for at least some cases. I'm not sure this should be merged into master, but on the other hand, master currently doesn't work properly.

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -46,9 +46,9 @@ type BYOCFindTilesDatasetParameters = {
 
 export class BYOCLayer extends AbstractSentinelHubV3Layer {
   public readonly dataset = DATASET_BYOC;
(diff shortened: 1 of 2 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -113,6 +113,10 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
         });
 
         this.locationId = res.data.location.id;
+      } else if (this.subType === BYOCSubTypes.ZARR) {
```
dgostencnik, line 116: Shouldn't this happen only if location is not set?

dgostencnik (approved):

PR #215, opened 2022-05-16
Compatible collections for order

dgostencnik:
>There is an not documented! as only for internal use endpoint that returns list of user's compatible collections. `api/v1/data_import/orders/searchcompatiblecollections` (you post your order detail and get back the list of user's compatible collections)

### DIFF
--- a/src/dataimport/const.ts
+++ b/src/dataimport/const.ts
@@ -103,6 +103,7 @@ export type TPDIOrderCompatibleCollection = {
   requiresMetadataUpdate: boolean;
   s3Bucket: string;
   userId: string;
+  additionalData?: Record<string, any>;
 };
 
 type LinksType = {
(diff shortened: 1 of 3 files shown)


src/dataimport/const.ts
```diff
@@ -95,6 +95,16 @@ export type TPDIOrderParams = {
   productKernel?: ResamplingKernel;
 };
 
+export type TPDIOrderCompatibleCollection = {
+  id: string;
```
Ardweaden, line 99: AIRBUS seems to also have `additionalData` field. I don't think it's very useful (like most of the fields), but it feels strange omitting a single field.

dgostencnik: I decided to remove everything except id and name as it is not being used

Ardweaden (approved):

PR #218, opened 2022-09-30
Feature/delete expired cache every minute

guss84:
- Removes expired cached items automatically every minute

### DIFF
--- a/src/utils/cacheHandlers.ts
+++ b/src/utils/cacheHandlers.ts
@@ -165,7 +165,7 @@ export const deleteExpiredCachedItemsAtInterval = (): void => {
   }, CLEAR_CACHE_INTERVAL);
 };
 
-export const findAndDeleteExpiredCachedItems = async (): Promise<void> => {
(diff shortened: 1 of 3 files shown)


src/utils/cacheHandlers.ts
```diff
@@ -155,6 +157,14 @@ const stringToHash = (message: string): number => {
...
 
+export const deleteExpiredCachedItemsAtInterval = (): void => {
+  // delete expired items on initialization
+  findAndDeleteExpiredCachedItems();
+  setInterval(() => {
+    findAndDeleteExpiredCachedItems();
+  }, CLEAR_CACHE_INTERVAL);
+};
+
 export const findAndDeleteExpiredCachedItems = async (): Promise<void> => {
```
zcernigoj, line 168: Nitpicking: This doesn't need to be exported since it's now used in only this file.

FFTR

zcernigoj (approved):

src/utils/cacheHandlers.ts
guss84: Removed the export.

PR #226, opened 2022-12-12
Add CreoDIAS collections

mitjak:


guss84 (commented):
Looks like it will be ok once https://github.com/sentinel-hub/sentinelhub-js/pull/228 is merged to this.

src/layer/S1GRDCDASLayer.ts
```diff
@@ -0,0 +1,6 @@
+import { DATASET_CDAS_S1GRD } from './dataset';
+import { S1GRDAWSEULayer } from './S1GRDAWSEULayer';
+
+export class S1GRDCDASLayer extends S1GRDAWSEULayer {
```
guss84, line 4: Do we know if `fis/getStats` is supported by CDAS, if not we should probably throw an error when the `getStats` method is called, since it is inherited from `AbstractSentinelHubV3Layer` .

`AbstractLayer` throws the following error

```
  public async getStats(payload: any, reqConfig?: RequestConfiguration): Promise<any> {
    throw new Error('getStats() not implemented for this dataset');
  }
```

mitjak: Makes sense, updated. Stats API should be supported eventually and we'll remove it then.

guss84: Should also throw an error for the other classes where `getStats` is not supported like `S2L1CCDASLayer` and `S2L2ACDASLayer `

guss84 (approved):

PR #240, opened 2023-04-21
Fix/use correct host for configuration service

dgostencnik:
```
  // Note that for SH v3 service, the endpoint for fetching the list of layers is always
  // https://services.sentinel-hub.com/, even for creodias datasets:
```
With CDSE  the assumption above is no longer correct.

### DIFF
--- a/src/layer/utils.ts
+++ b/src/layer/utils.ts
@@ -2,7 +2,12 @@ import axios, { AxiosRequestConfig } from 'axios';
 import { stringify, parseUrl, stringifyUrl } from 'query-string';
 import { parseStringPromise } from 'xml2js';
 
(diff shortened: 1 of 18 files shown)


src/layer/utils.ts
```diff
@@ -140,17 +145,33 @@ export function parseSHInstanceId(baseUrl: string): string {
...
+
+  if (/\ogc\/wms/.test(baseUrl)) {
+    host = baseUrl.substring(0, baseUrl.indexOf('/ogc/wms') + 1);
+  }
+
+  if (/dataspace.copernicus.eu/.test(host)) {
+    return host;
+  }
+
+  // Note that for SH v3 service, the endpoint for fetching the list of layers is always
```
mitjak, line 159: Since this comment is not correct in general anymore we should perhaps mention the exception of Copernicus endpoint.

PR #245, opened 2023-06-22
Feature/remove eo cloud

mitjak:
https://git.sinergise.com/team-6/eobrowser3/-/merge_requests/1337


dgostencnik:
As I said, I'm not sure if these layers should be removed from sh-js just because they are no longer used by eob. Maybe it would be less radical to throw an error message like "This layer is deprecated" when you try to create eocloud layers.

One way or another, we can use this opportunity to stop abusing patch versions for major changes

dgostencnik (approved):

PR #246, opened 2023-06-28
Proba v update

mitjak:
https://git.sinergise.com/team-6/eobrowser3/-/issues/1040


src/layer/wms.ts
```diff
@@ -89,16 +89,16 @@ export function wmsGetMapUrl(
   }
 
   if (!params.fromTime) {
-    queryParams.time = moment.utc(params.toTime).format('YYYY-MM-DDTHH:mm:ss') + 'Z';
+    queryParams.time = moment.utc(params.toTime).format('YYYY-MM-DD');
```
dgostencnik, line 92: why was this changed to date only?

mitjak: Proba-V doesn't support time in single date mode. Setting this universally makes sense as we're actually searching for date and not datetime in this case. Other providers support this as well.

dgostencnik: I think this should be handled by client (eo-browser in this case) and not by library. WMS handler should be agnostic and  doesn't know anything about proba-v or any other wms service. Just because eo-browser uses some wms service which doesn't support time,  that doesn't mean that 
- all wms services supports date only
- no one will ever want to set time, after all parameter is called `time` and not date

Personally, I would do this in eob

mitjak: I generally agree. But am also not sure EOB is the right place to handle these peculiarities. All in all the right way of addressing these would be to read getCapabilities more rigorously and then act accordingly (not sure if it does indeed contain all the required info). But am not sure if this is something worth pursuing at this point. 

Perhaps I would rather suggest creating a separate layer for the WMT MS and handle it there. What do you think?

dgostencnik: Yes, I think this could work.

src/layer/LayersFactory.ts
```diff
@@ -332,6 +340,36 @@ export class LayersFactory {
     );
   }
 
+  private static async makeLayersWmsWmtMs(
```
dgostencnik, line 343: This could be merged with `makeLayersWms`as  it is duplicated except for the layer class.  I think one function with additional param layerClass would be better. 

Or perhaps even better option would be to  change  `fetchLayersFromGetCapabilitiesXml` a bit (extract parsing capabilities into own function) and use `fetchGetCapabilitiesXml` to get capabilities and determine with isWMSCapabilities which layer should be created. In this scenario we could also remove hardcoded url  above (services.terrascope.be/wms/v2)

dgostencnik (approved):

PR #247, opened 2023-07-03
Add mosaicking order to S1GRD constructor

zansinergise:
https://git.sinergise.com/team-6/eobrowser3/-/issues/607

Constructor was missing the parameter `mosaickingOrder` so it was always set to null.

### DIFF
--- a/src/layer/S1GRDAWSEULayer.ts
+++ b/src/layer/S1GRDAWSEULayer.ts
@@ -12,6 +12,7 @@ import {
   LinkType,
   DataProductId,
   FindTilesAdditionalParameters,
+  MosaickingOrder,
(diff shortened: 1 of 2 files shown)


dgostencnik (commented):
I know that tests for constructor parameters are missing, but it would be nice if you could also add test for this one (just mosaickingOrder parameter).  fftr

zansinergise:
> I know that tests for constructor parameters are missing, but it would be nice if you could also add test for this one (just mosaickingOrder parameter). fftr

Added 1 test for creating layer without mosaickingOrder parameter and 1 test with mosaickingOrder parameter.

dgostencnik (approved):

PR #253, opened 2023-10-09
Add a setter function for tpdi service url

mitjak:


dgostencnik:
Perhaps one test for changing tpdi service url could be also added to existing `Test TPDI service enpoints` tests in `/dataimport/__tests__/TPDI.ts`

mitjak:
Agree, makes total sense. Added tests now as well.

dgostencnik (approved):

PR #259, opened 2024-01-29
Feature/low res byoc

mitjak:
Low-resolution collection support in BYOC layer

https://git.sinergise.com/team-6/eobrowser3/-/merge_requests/1623

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -147,17 +147,21 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
   ): Promise<ProcessingPayload> {
     await this.updateLayerFromServiceIfNeeded(reqConfig);
 
-    if (
-      this.lowResolutionCollectionId !== undefined &&
-      this.lowResolutionMetersPerPixelThreshold !== undefined &&
-      metersPerPixel(params.bbox, payload.output.width) > this.lowResolutionMetersPerPixelThreshold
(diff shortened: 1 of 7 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -220,6 +237,9 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
   }
 
   protected getCatalogCollectionId(): string {
+    if (this.lowResolutionCollectionId !== undefined) {
```
dgostencnik, line 240: Not sure about this. I suppose it will work as expected as long as low res collection in derived from high res collection.

dgostencnik (commented):
I think tests for switching collections should also be added

src/layer/__tests__/BYOCLayer.ts
```diff
@@ -287,3 +287,104 @@ describe.only('shServiceRootUrl', () => {
...
   });
 });
+
+describe.only('shServiceRootUrl', () => {
+  beforeEach(async () => {
+    setAuthToken(AUTH_TOKEN);
+    mockNetwork.reset();
+  });
+
+  test.each([
```
dgostencnik, line 297: I would also include use case when  `lowResolutionCollectionId` or `lowResolutionMetersPerPixelThreshold` is not set. fftr

dgostencnik (approved):

PR #265, opened 2024-04-25
Add lowest cc to flyovers

rokrak1:
Added propetry lowestCloudCover flyover meta data.

### DIFF
--- a/src/layer/AbstractLayer.ts
+++ b/src/layer/AbstractLayer.ts
@@ -310,7 +310,9 @@ export class AbstractLayer {
               fromTime: tiles[tileIndex].sensingTime,
               toTime: tiles[tileIndex].sensingTime,
               coveragePercent: 0,
-              meta: {},
+              meta: {
+                lowestCloudCoverPercent: tiles[tileIndex].meta.cloudCoverPercent,
+              },
             };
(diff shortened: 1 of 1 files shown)


src/layer/AbstractLayer.ts
```diff
@@ -356,6 +360,12 @@ export class AbstractLayer {
               this.roundCoordinates(currentFlyoverGeometry),
               this.roundCoordinates(tiles[tileIndex].geometry.coordinates as Geom),
             );
+
+            // update the lowest cloud cover:
+            if (tiles[tileIndex].meta.cloudCoverPercent < tiles[tileIndex - 1].meta.cloudCoverPercent) {
```
zcernigoj, line 365: I think this check would be clearer if the comparison would be done against the saved value and not the value of the previous element 

```js
if (tiles[tileIndex].meta.cloudCoverPercent < flyovers[flyoverIndex].meta.lowestCloudCoverPercent) {
  ...
}
```

zcernigoj, line 365: I think this check would be clearer if the comparison would be done against the saved value and not the value of the previous element 

```js
if (tiles[tileIndex].meta.cloudCoverPercent < flyovers[flyoverIndex].meta.lowestCloudCoverPercent) {
  ...
}
```

zcernigoj (changes requested):

PR #273, opened 2024-08-02
Replace xml2js with fast-xml-parser

janybravo:
To remove dependency/peerDependency of xml2js that causes problems with VITE. Backwards compatibility with xml2js is preserved by using XmlParserOptions export with API of fast-xml-parse.

### DIFF
--- a/.github/workflows/on_push.yaml
+++ b/.github/workflows/on_push.yaml
@@ -21,6 +21,7 @@ jobs:
         run: |
           npm ci
           npm run lint
+          npm run build
 
       - name: Run tests
         run: |
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,9 @@
       "engines": {
         "node": ">=18"
       },
+      "optionalDependencies": {
+        "@rollup/rollup-linux-x64-gnu": "^4.19.2"
+      },
       "peerDependencies": {
         "@turf/area": "^6.0.1",
         "@turf/helpers": "^6.1.4",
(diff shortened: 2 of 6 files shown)


zcernigoj:
As far as I tested, everything works as before. But I would also ask someone else to additionally check.

The things I found out that also don't work on master (published version):

1: Time dimension is in a different place in SH WMS GetCapabilities response - it's not in every layer, but under the parent `Layer` element in the XML. 
So, the `WMSLayer.findDatesUTC` fails, because it assumes that the time dimension is in every layer (https://github.com/sentinel-hub/sentinelhub-js/blob/master/src/layer/WmsLayer.ts#L69). Probably best to solve in a different MR.

2: `WMTSLayer` doesn't have the `findDates` implemented, so we "can't test" if the dates are parsed correctly from the WMTS GetCapabilities response.
Getting dates from WMTS GetCapabilities response was implemented in EO Browser directly in the past, and now no WMTS datasources are used, so not needed from our side. Might be useful for others though.

3: The build process fails for me (on linux) with the error below and the mentioned solution (removing `node_modules` and `package-lock.json`) works. 
The problem is that after installing dependencies again, the `package-lock.json` is totally different from before. (happens for both node 18 and 20)

```
.../sentinelhub-js(update-dependencies2)$ npm run build

> @sentinel-hub/sentinelhub-js@1.0.0 build
> rollup -c

.../sentinelhub-js/node_modules/rollup/dist/native.js:59
                throw new Error(
                ^

Error: Cannot find module @rollup/rollup-linux-x64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both package-lock.json and node_modules directory.
    at requireWithFriendlyError (.../sentinelhub-js/node_modules/rollup/dist/native.js:59:9)
    at Object.<anonymous> (.../sentinelhub-js/node_modules/rollup/dist/native.js:68:76)
    ... 3 lines matching cause stack trace ...
    at Module._load (node:internal/modules/cjs/loader:1013:12)
    at Module.require (node:internal/modules/cjs/loader:1225:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (.../sentinelhub-js/node_modules/rollup/dist/shared/parseAst.js:12:19)
    at Module._compile (node:internal/modules/cjs/loader:1356:14) {
  [cause]: Error: Cannot find module '@rollup/rollup-linux-x64-gnu'
  Require stack:
  - .../sentinelhub-js/node_modules/rollup/dist/native.js
  - .../sentinelhub-js/node_modules/rollup/dist/shared/parseAst.js
  - .../sentinelhub-js/node_modules/rollup/dist/shared/rollup.js
  - .../sentinelhub-js/node_modules/rollup/dist/bin/rollup
      at Module._resolveFilename (node:internal/modules/cjs/loader:1134:15)
      at Module._load (node:internal/modules/cjs/loader:975:27)
      at Module.require (node:internal/modules/cjs/loader:1225:19)
      at require (node:internal/modules/helpers:177:18)
      at requireWithFriendlyError (.../sentinelhub-js/node_modules/rollup/dist/native.js:41:10)
      at Object.<anonymous> (.../sentinelhub-js/node_modules/rollup/dist/native.js:68:76)
      at Module._compile (node:internal/modules/cjs/loader:1356:14)
      at Module._extensions..js (node:internal/modules/cjs/loader:1414:10)
      at Module.load (node:internal/modules/cjs/loader:1197:32)
      at Module._load (node:internal/modules/cjs/loader:1013:12) {
    code: 'MODULE_NOT_FOUND',
    requireStack: [
      '.../sentinelhub-js/node_modules/rollup/dist/native.js',
      '.../sentinelhub-js/node_modules/rollup/dist/shared/parseAst.js',
      '.../sentinelhub-js/node_modules/rollup/dist/shared/rollup.js',
      '.../sentinelhub-js/node_modules/rollup/dist/bin/rollup'
    ]
  }
}

Node.js v18.19.0
```

PR #275, opened 2024-08-19
Fix - S3 OLCI L2 extends layer with CC

zansinergise:
https://hello.planet.com/code/sentinel-hub/sentinel-frontend/cdse/copernicus-browser/-/issues/208

S3 OLCI does not support cloud coverage, but S3 OLCI L2 does, so I just extended the class with AbstractSentinelHubV3WithCCLayer and added some missing functions specific to S3 OLCI

### DIFF
--- a/src/layer/S3OLCIL2CDASLayer.ts
+++ b/src/layer/S3OLCIL2CDASLayer.ts
@@ -1,6 +1,18 @@
 import { DATASET_CDAS_S3OLCIL2 } from './dataset';
 import { S3SLSTRCDASLayer } from './S3SLSTRCDASLayer';
 
+import { Link, LinkType } from './const';
+
 export class S3OLCIL2CDASLayer extends S3SLSTRCDASLayer {
   public readonly dataset = DATASET_CDAS_S3OLCIL2;
+
+  protected getTileLinksFromCatalog(feature: Record<string, any>): Link[] {
+    const { assets } = feature;
+    let result: Link[] = super.getTileLinksFromCatalog(feature);
+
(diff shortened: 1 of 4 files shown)


janybravo:
Intention of this fix would appear clearer if all commits are squashed into one with message eg. "Fix S3 OLCI L2 for Cloud Coverage to work properly" (+rebased on master). That would possibly help readers tiny bit reading such git message 😃.

src/layer/S3OLCIL2CDASLayer.ts
```diff
@@ -1,6 +1,18 @@
 import { DATASET_CDAS_S3OLCIL2 } from './dataset';
-import { S3OLCILayer } from './S3OLCILayer';
+import { S3SLSTRCDASLayer } from './S3SLSTRCDASLayer';
 
-export class S3OLCIL2CDASLayer extends S3OLCILayer {
+import { Link, LinkType } from './const';
+
+export class S3OLCIL2CDASLayer extends S3SLSTRCDASLayer {
```
guss84, line 6: We shouldnt extend from `S3SLSTRCDASLayer`, as S3-SLSTR has different properties than S3-OLCI, for example `S3SLSTRView`. It might be best to just extend from `AbstractSentinelHubV3WithCCLayer`.

guss84 (changes requested):

src/layer/S3OLCIL2CDASLayer.ts
zansinergise: Agreed and changed

guss84 (approved):

PR #278, opened 2024-09-04
Change AbstractSentinelHubV3Layer to receive evalscript value from dataProduct endpoint

janybravo:
Based on https://hello.planet.com/code/sentinel-hub/sentinel-frontend/eo-browser/eobrowser3/-/issues/1510

### DIFF
--- a/src/layer/AbstractSentinelHubV3Layer.ts
+++ b/src/layer/AbstractSentinelHubV3Layer.ts
@@ -28,11 +28,11 @@ import { wmsGetMapUrl } from './wms';
 
 import { Effects } from '../mapDataManipulation/const';
 import { runEffectFunctions } from '../mapDataManipulation/runEffectFunctions';
+import { StatisticsProviderType } from '../statistics/const';
 import { Fis } from '../statistics/Fis';
 import { StatisticalApi } from '../statistics/StatisticalApi';
 import { CACHE_CONFIG_30MIN, CACHE_CONFIG_NOCACHE } from '../utils/cacheHandlers';
-import { fetchLayerParamsFromConfigurationService, getSHServiceRootUrl } from './utils';
-import { StatisticsProviderType } from '../statistics/const';
+import { fetchDataProduct, fetchLayerParamsFromConfigurationService, getSHServiceRootUrl } from './utils';
 
 interface ConstructorParameters {
   instanceId?: string | null;
@@ -127,6 +127,19 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
     if (!layerParams) {
       throw new Error('Layer params could not be found');
     }
+
+    if (
+      !layerParams['evalscript'] &&
+      layerParams['dataProduct'] &&
+      !SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(layerParams['dataProduct'])
+    ) {
+      const response = await fetchDataProduct(layerParams['dataProduct'], reqConfig);
+      const evalScript = response?.data?.evalScript;
+      if (evalScript) {
+        layerParams['evalscript'] = evalScript;
+      }
+    }
+
     return layerParams;
   }
 
@@ -236,10 +249,10 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   }
 
   public supportsApiType(api: ApiType): boolean {
-    if (this.dataProduct && !SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(this.dataProduct)) {
-      return api === ApiType.WMS;
+    if (this.dataProduct && SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(this.dataProduct)) {
+      return api === ApiType.PROCESSING;
     }
-    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset);
+    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset && !!this.evalscript);
   }
 
   protected getWmsGetMapUrlAdditionalParameters(): Record<string, any> {
@@ -679,7 +692,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       if (!this.downsampling && layerParams.downsampling) {
         this.downsampling = layerParams.downsampling;
       }
-      // this is a hotfix for `supportsApiType()` not having enough information - should be fixed properly later:
+
       this.dataProduct = layerParams['dataProduct'] ? layerParams['dataProduct'] : null;
     }, reqConfig);
   }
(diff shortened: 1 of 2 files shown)


src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -668,8 +668,16 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
       this.legend = layerParams['legend'] ? layerParams['legend'] : null;
 
       if (!this.evalscript) {
-        this.evalscript = layerParams['evalscript'] ? layerParams['evalscript'] : null;
+        const dataProductUrl = layerParams['dataProduct'];
+        this.evalscript = layerParams['evalscript'];
+        if (!this.evalscript && dataProductUrl) {
```
zcernigoj, line 673: I think checking if evalscript is present and fetching the evalscript could be directly in `fetchLayerParamsFromSHServiceV3()` which is also called in `getMap()` if `evalscript` or `dataProduct` are not defined. This would in some cases eliminate the need for usage of `layer.updateLayerFromServiceIfNeeded()` in the apps.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -668,8 +668,16 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
+          const response = await fetchDataProduct(dataProductUrl, innerReqConfig);
+          this.evalscript = response.data?.evalScript;
+          // now that evalscript here, dataProduct=null will make rest of logic use ApiType.PROCESSING
+          this.dataProduct = null;
```
zcernigoj, line 677: I think setting `this.dataProduct` to `null` is not needed, because `createProcessingPayload()` used in `getMap()` checks if evalscript is set first (and uses it (instead of `dataProduct`) if it is).

zcernigoj (changes requested):
I think a release candidate version can already be created and a MR prepared in EO Browser repository with this new release candidate version.

src/layer/AbstractSentinelHubV3Layer.ts
janybravo: createProcessingPayload is not called if apiType is determined wms.

src/layer/AbstractSentinelHubV3Layer.ts
janybravo: Moved, thanks!

src/layer/AbstractSentinelHubV3Layer.ts
zcernigoj: True, in that case the `getMap` from the parent class is called, but it does not use `dataProduct` so it's not important if the value is kept or not. Let's leave it the way you did it.

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -127,6 +127,19 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
...
       throw new Error('Layer params could not be found');
     }
+
+    const dataProductUrl = layerParams['dataProduct'];
+
+    if (!this.evalscript && dataProductUrl) {
+      const response = await fetchDataProduct(dataProductUrl, reqConfig);
+      const evalScript = response?.data?.evalScript;
+      if (evalScript) {
+        this.evalscript = evalScript;
```
zcernigoj, line 137: Would it make sense to set evalscript to `layerParams` object that is then returned and not to the object?

Every piece of code where this function is called, then deals with setting the evalscript to the object if it's not set yet.

True, with this approach, setting `dataProduct` to `null` would need to be added to more places.

But I think it's worth it - to not pollute the function that is responsible only for fetching the data with setting object properties (following separation of concerns).

zcernigoj (changes requested):

src/layer/AbstractSentinelHubV3Layer.ts
janybravo: Maybe I could unwind this logic a bit then

src/layer/AbstractSentinelHubV3Layer.ts
```diff
@@ -236,7 +245,7 @@ export class AbstractSentinelHubV3Layer extends AbstractLayer {
   }
 
   public supportsApiType(api: ApiType): boolean {
-    if (this.dataProduct && !SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(this.dataProduct)) {
+    if (!this.evalscript || SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(this.dataProduct)) {
```
zcernigoj, line 248: Sorry for so many complications in this MR

I checked and the combination of the evalscript being successfully fetched and set for the data products in `SUPPORTED_DATA_PRODUCTS_PROCESSING` in `fetchLayerParamsFromSHServiceV3` causes problems, because the evalscript is set and JS doesn't even check the condition after `||`, so it uses ProcessAPI. Additionally, even if the WMS ApiType is selected, it will still use the evalscript instead of the data product.

I realized that `dataProduct` can also be passed to ProcessAPI - and this is already done in `src/layer/processing.ts` in case the `dataProduct` parameter is defined and `evalscript` isn't. ([lines 156 to 161](https://github.com/sentinel-hub/sentinelhub-js/blob/master/src/layer/processing.ts#L156))

I tweaked the code a bit, so it works also for the data product in `SUPPORTED_DATA_PRODUCTS_PROCESSING`  (for me locally):

- changed the new `if` in `fetchLayerParamsFromSHServiceV3` so that it checks if the `dataProduct` is in `SUPPORTED_DATA_PRODUCTS_PROCESSING`, so the evalscript is not requested if it is

```js
if (!layerParams['evalscript'] && layerParams['dataProduct'] && !SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(layerParams['dataProduct'])) {
  // call fetchDataProduct() ... 
}
```

- changed the checks in this function (`supportsApiType`) entirely

```js
  public supportsApiType(api: ApiType): boolean {
    if (this.dataProduct && SUPPORTED_DATA_PRODUCTS_PROCESSING.includes(this.dataProduct)){
      return api === ApiType.PROCESSING;
    }
    return api === ApiType.WMS || (api === ApiType.PROCESSING && !!this.dataset && !!this.evalscript);
  }
```

The changed tests can be changed back

zcernigoj (changes requested):

src/layer/AbstractSentinelHubV3Layer.ts
janybravo: Wasn't aware of dataProduct input parameter from processing API. Thanks for such insights!

zcernigoj (approved):
Looks good now, thanks!

PR #294, opened 2025-01-21
add fcm3m and fcd v1.2

dthiex:
Same as https://github.com/sentinel-hub/sentinelhub-js/pull/285 which was merged by accident and reverted.


dthiex:
@mitjak Can you approve this again? It's the same as https://github.com/sentinel-hub/sentinelhub-js/pull/285 which I merged a bit to quickly yesterday and then reverted. I wanted to keep main clean but got today the confirmation that we can move on with this.

mitjak (approved):

Issue #300, opened 2025-06-06
Replace deprecated /metadata/collection/:id to use with newly added staging1 serevices hostname


janybravo:
Original [GitLab issue](See https://hello.planet.com/code/sentinel-hub/sentinel-frontend/pip-browser/-/issues/226).

PR #301, opened 2025-06-06
Replace deprecated /api/v1/metadata/collection/:id

janybravo:
Closes https://github.com/sentinel-hub/sentinelhub-js/issues/300

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -114,21 +114,10 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       }
 
       if (this.locationId === null) {
-        if (this.subType !== BYOCSubTypes.ZARR) {
-          const url = `${this.getSHServiceRootUrl()}api/v1/metadata/collection/${this.getTypeId()}`;
-          const headers = { Authorization: `Bearer ${getAuthToken()}` };
-          const res = await axios.get(url, {
-            responseType: 'json',
-            headers: headers,
-            ...getAxiosReqParams(innerReqConfig, CACHE_CONFIG_30MIN),
-          });
-
-          this.locationId = res.data.location.id;
-        } else {
-          // Obtaining location ID is currently not possible for ZARR.
-          // We hardcode AWS EU as the only currently supported location.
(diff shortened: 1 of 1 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -304,17 +293,17 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
...
-
-      // we need bands' sampleTypes that are returned from Metadata API
-      // but want to use only bands that are also returned from Catalog API
-      return metadataBands.filter((mb) => !!catalogBands.find((cb) => cb.name === mb.name));
+      const metadataBands: BYOCBand[] = (
+        metadataRes.data.summaries['eo:bands'] as { name: string }[] | undefined
+      )?.map(({ name }, index) => {
+        return {
+          name,
+          sampleType: metadataRes.data.summaries?.['raster:bands'][index].data_type?.toUpperCase(),
```
zcernigoj, line 303: This assumes that all elements in `eo:bands` have a "sibling" in `raster:bands` at exactly the same index.

I haven't found a collection where this wouldn't be true (PSD collections), but it would be best to check.

src/layer/const.ts
```diff
@@ -149,6 +149,8 @@ export type ImageProperties = {
 
 export const SH_SERVICE_HOSTNAMES_V3: string[] = [
   'https://services.sentinel-hub.com/',
+  'https://staging1-awseu.sentinel-hub.com/',
+  'https://staging0-awseu.sentinel-hub.com/',
```
zcernigoj, line 153: Will check with the team if it's ok to have these staging URLs public.

zcernigoj: In general the `registerHostnameReplacing(originalHostname, desiredHostname)` could be used to "override" the all urls with `originalHostname` with the `desiredHostname`, but the underlying `replaceHostnamesInterceptor` doesn't seem to work properly - checking.

janybravo: Agree, removed those extra urls and let registerHostnameReplacing handle its work.

src/layer/BYOCLayer.ts
zcernigoj: Great, thanks.

PR #302, opened 2025-07-16
Location support fix

mitjak:
- re-enables reading location id for BYOC from the API
- updates GCP service location url

https://hello.planet.com/code/sentinel-hub/sentinel-frontend/pip-browser/-/issues/238

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -114,10 +114,21 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       }
 
       if (this.locationId === null) {
-        this.locationId =
-          (Object.entries(SHV3_LOCATIONS_ROOT_URL).find(
-            ([, url]) => url === this.shServiceRootUrl,
-          )?.[0] as LocationIdSHv3) ?? LocationIdSHv3.awsEuCentral1;
+        if (this.subType !== BYOCSubTypes.ZARR) {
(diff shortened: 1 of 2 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -114,10 +114,21 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       }
 
       if (this.locationId === null) {
-        this.locationId =
-          (Object.entries(SHV3_LOCATIONS_ROOT_URL).find(
-            ([, url]) => url === this.shServiceRootUrl,
-          )?.[0] as LocationIdSHv3) ?? LocationIdSHv3.awsEuCentral1;
+        if (this.subType !== BYOCSubTypes.ZARR) {
+          const url = `${this.getSHServiceRootUrl()}api/v1/byoc/global/?ids=${this.collectionId}`;
```
zcernigoj, line 118: This endpoint probably doesn't return batch collections (undocumented, so we can't know), so I would change the if to `if (this.subType === BYOCSubTypes.BYOC) {` just for peace of mind.

Do you have more info about this endpoint?

mitjak: I double checked and it's ok to use this endpoint for all types. It even supports ZARRs, but I wouldn't change the logic here just yet as there might be other dependencies for those.

zcernigoj (approved):

PR #312, opened 2026-01-26
fix matchDatasetFromGetCapabilities so it returns correct Dataset object for Landsat on AWS

zcernigoj:
When creating layers (`_Layer` objects) with `LayersFactory.makeLayers(baseUrl)`, the provided `baseUrl` is a URL for the _configuration_.

This causes problems when a _configuration_ contains layers that use _collection_ from a different deployment of SH. e.g. configuration on SH @ AWS EU Central 1 contains a layer that uses Landsat collection, which is only on AWS US West 2.

Configurations that are on AWS EU Central 1 can contain layers using collections from AWS US West 2.
But using the baseUrl from the configuration won't find correct Dataset object.
  Thus the apps using sentinelhub.js fail when preparing list of collections for the selected configuration.

Workaround is to check [if the `baseUrl` matches _AWS EU Central 1_] and [if the `shServiceHostname` from the datasets with `shJsonGetCapabilitiesDataset` matching the provided `datasetId` matches _AWS US West 2_].

In both checks above are true, an appropriate dataset is found.

### DIFF
--- a/src/layer/LayersFactory.ts
+++ b/src/layer/LayersFactory.ts
@@ -11,7 +11,13 @@ import {
   isWMSCapabilities,
 } from './utils';
 import { ensureTimeout } from '../utils/ensureTimeout';
-import { OgcServiceTypes, SH_SERVICE_HOSTNAMES_V3, PLANET_FALSE_COLOR_TEMPLATES } from './const';
+import {
+  OgcServiceTypes,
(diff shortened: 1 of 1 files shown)


rokrak1 (approved):
Looks good 👍

zansinergise (approved):

PR #313, opened 2026-02-04
Export needed functions

zansinergise:
Just an export of 2 functions which would be needed for CDSE FE support of applying effects


zcernigoj (approved):
Types in those functions could be improved 
- `getImageProperties()` has a couple of `any`
- Cursor really wants to change `const newImg = new ImageData(rgba, width, height);` to `const newImg = new ImageData(new Uint8ClampedArray(rgba), width, height);` in `getBlob()`

Not critical though.

PR #319, opened 2026-06-15
Improve error handling for BYOCLayer

zcernigoj:
Added checks in `BYOCLayer` that throw errors
- if `shServiceRootUrl` is not set
- if response from `api/v1/byoc/global/?ids=coll-id` is empty
  - can mean the user doesn't have access to the collection (anymore)

### DIFF
--- a/src/layer/BYOCLayer.ts
+++ b/src/layer/BYOCLayer.ts
@@ -115,7 +115,13 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
 
       if (this.locationId === null) {
         if (this.subType !== BYOCSubTypes.ZARR) {
-          const url = `${this.getSHServiceRootUrl()}api/v1/byoc/global/?ids=${this.collectionId}`;
+          if (!this.shServiceRootUrl) {
+            const err = 'Service URL not found.';
+            console.error(err);
+            throw new Error(err);
+          }
+
+          const url = `${this.shServiceRootUrl}api/v1/byoc/global/?ids=${this.collectionId}`;
           const headers = { Authorization: `Bearer ${getAuthToken()}` };
           const res = await axios.get(url, {
             responseType: 'json',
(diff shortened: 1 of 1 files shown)


src/layer/BYOCLayer.ts
```diff
@@ -115,15 +115,28 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
 
       if (this.locationId === null) {
         if (this.subType !== BYOCSubTypes.ZARR) {
-          const url = `${this.getSHServiceRootUrl()}api/v1/byoc/global/?ids=${this.collectionId}`;
+          if (!this.shServiceRootUrl) {
+            const err = 'Service URL not found.';
+            console.error(err);
```
guss84: Do we want to also log the errors while throwing them? We are also not consistent in logging errors.

src/layer/BYOCLayer.ts
```diff
@@ -284,14 +297,17 @@ export class BYOCLayer extends AbstractSentinelHubV3Layer {
       if (this.subType === BYOCSubTypes.ZARR) {
         throw new Error('Fetching available bands for ZARR not supported.');
       }
+      if (!this.shServiceRootUrl) {
+        throw new Error('Service URL not set.');
```
guss84: Should we change this msg to `Service URL not found.` to be consistent with the message above?

src/layer/BYOCLayer.ts
zcernigoj: > We are also not consistent in logging errors.

The inconsistency is the main reason why I also added `console.error`. Sometimes we also log them along with throwing, sometimes we don't. So, I think we should decide and try to follow it going forward.

Since this is a library, I think it's nicer if we only throw them, and the app developers (which is also us) can decide if they want to log those errors (and potentially also display them in the UI).

src/layer/BYOCLayer.ts
zcernigoj: Good point, I think we should. Fixed.

src/layer/BYOCLayer.ts
guss84: yes I agree it would be better to just throw the errors

zcernigoj: Removed

